Issue #8, opened 2022-03-16
Add owlracle (gas estimate) integration.


groninge01:

0xSkly:
closing due to inactivity

PR #9, opened 2022-03-17
Add owlracle api

groninge01:
Adds the owlracle gas api into the backend.


groninge01:
Still working on it...

groninge01:
@danielmkm This PR has been updated and is ready for review now.

groninge01:
Fixes #8.

app/env.ts
```diff
@@ -30,6 +30,7 @@ export const schema = {
     COPPER_PROXY_ADDRESS: String,
     RPC_URL: String,
     YEARN_VAULTS_ENDPOINT: String,
+    OWLRACLE_API_KEY: String,
```
danielmkm, line 33: @franzns can you make sure you add the api key to all deployment

modules/gas-estimate/lib/owlracle.service.ts
```diff
@@ -0,0 +1,55 @@
...
+    percentile: number;
+    accept: string;
+    version: number;
+}
+export class OwlracleService {
+    private readonly baseUrl: string;
+    private readonly requestParams: OwlracleRequest;
+
+    constructor() {
+        this.baseUrl = 'https://owlracle.info/ftm/gas';
```
danielmkm, line 18: make this url configurable, ie: what if this needed to work for a different chain?

app/env.ts
franzns: I can do that, if you can get it to me :)

franzns: env var is configured everywhere

modules/gas-estimate/lib/owlracle.service.ts
groninge01: @danielmkm I have updated the PR to make it multi chain. If it's ok I'll need to update the FE implementation too.

0xSkly:
closed due to inactivity, we can bring it back later

PR #10, opened 2022-03-18
Edit history refresh triggers

franzns:
I'm not 100% sure I got this correctly but AFAIK:

- When a request asks for portfolio history it checks the redis cache, if it's not in there, it will create the history from the sql db and put it into the redis cache, but doesn't refresh refreshLatestBlockCachedTimestamp() which I think is wrong.
- On the other hand, the cron writes everything into the sql db and triggers the refreshLatestBlockCachedTimestamp() without putting anything into redis, which does not make sense.

TBH: not sure I got everything, but this might be why we have such strange behaviour.


franzns:
Hm, nevermind, is correct nevertheless :D

PR #13, opened 2022-03-28
Fix zap portfolio

### LINKED ISSUE #2: Zapping into a new farm is not reflected in My Portfolio page
To re-create the issue:
- Invest in a pool, where the user has not invested before
- Use the ZAP feature

Outcome:
The new investment shows up under "My Investment" but is not reflected under "Pools".
If it is a new asset, it is not reflected under "Assets" either. If the portfolio contained the asset before, it is reflected under "Assets".

Re-created the issue using wallet `0x500f8fe004e978d1c58dba852bfe7bba8ae434c1` that joined pool `0xc9ba718a71bfa34cedea910ac37b879e5913c14e0002000000000000000001ad` and ZAPped into the farm `42`.
The user is correctly added to the masterchef subgraph, as seen in this query: [graphql](https://graph-node.beets-ftm-node.com/subgraphs/name/masterchef/graphql?query=%7B%0A%20%20pools(where%3A%20%7Bid%3A%20%2242%22%7D)%20%7B%0A%20%20%20%20id%0A%20%20%20%20pair%0A%20%20%20%20userCount%0A%20%20%20%20users(where%3A%20%7Bid%3A%20%2242-0x500f8fe004e978d1c58dba852bfe7bba8ae434c1%22%7D)%20%7B%0A%20%20%20%20%20%20id%0A%20%20%20%20%7D%0A%20%20%7D%0A%7D%0A)

franzns:
Need to also get the pools from the Masterchef because if I zap into a farm directly without ever joined the pool, I never owned the BPT and will not be part of the sharesOwned query in the balancer subgraph.

We will get the pool address and pool id from Masterchef subgraph and balancer subgraph respectively, loop through all pools and match with the pool id/address. Need to manually construct a sharesOwned type, especially needed for pool that are only in Masterchef .

fixes #2 

I assume there is a nicer typescript way to construct the pool-id/address array :) help!

### DIFF
--- a/modules/portfolio/lib/portfolio-data.service.ts
+++ b/modules/portfolio/lib/portfolio-data.service.ts
@@ -92,10 +92,12 @@ export class PortfolioDataService {
         const blockNumber = parseInt(block.number);
         const poolShares = [];
 
-        const poolIds = [];
+        const poolIds: string[] = [];
+        const poolAddresses: string[] = [];
+
         for (const farm of farmUsers) {
             if (farm.pool) {
-                poolIds.push(farm.pool?.pair);
+                poolAddresses.push(farm.pool.pair);
             }
         }
         for (const sharesOwned of user.sharesOwned ?? []) {
@@ -103,8 +105,7 @@ export class PortfolioDataService {
         }
 
         for (const pool of pools) {
-            // pool.address for masterchef subgraph, pool.id for sharesOwned in balancer subgraph
-            if (poolIds.includes(pool.address) || poolIds.includes(pool.id)) {
+            if (poolAddresses.includes(pool.address) || poolIds.includes(pool.id)) {
                 const poolSnapshot: PrismaBalancerPoolSnapshotWithTokens = {
                     ...pool,
                     poolId: pool.id,
--- a/modules/portfolio/lib/portfolio-data.service.ts
+++ b/modules/portfolio/lib/portfolio-data.service.ts
@@ -92,35 +92,50 @@ export class PortfolioDataService {
         const blockNumber = parseInt(block.number);
         const poolShares = [];
 
-        for (const sharesOwned of user.sharesOwned ?? []) {
-            const pool = pools.find((pool) => pool.id === sharesOwned.poolId.id);
-
-            if (!pool) {
-                continue;
+        const poolIds = [];
+        for (const farm of farmUsers) {
+            if (farm.pool) {
+                poolIds.push(farm.pool?.pair);
             }
+        }
+        for (const sharesOwned of user.sharesOwned ?? []) {
+            poolIds.push(sharesOwned.poolId.id);
+        }
 
-            const poolSnapshot: PrismaBalancerPoolSnapshotWithTokens = {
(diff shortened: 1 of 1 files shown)


modules/portfolio/lib/portfolio-data.service.ts
```diff
@@ -92,35 +92,50 @@ export class PortfolioDataService {
...
 
-        for (const sharesOwned of user.sharesOwned ?? []) {
-            const pool = pools.find((pool) => pool.id === sharesOwned.poolId.id);
-
-            if (!pool) {
-                continue;
+        const poolIds = [];
+        for (const farm of farmUsers) {
+            if (farm.pool) {
+                poolIds.push(farm.pool?.pair);
```
danielmkm, line 98: you don't need the `?` here since you check the presence in your if statement above

modules/portfolio/lib/portfolio-data.service.ts
```diff
@@ -92,35 +92,50 @@ export class PortfolioDataService {
         const blockNumber = parseInt(block.number);
         const poolShares = [];
 
-        for (const sharesOwned of user.sharesOwned ?? []) {
-            const pool = pools.find((pool) => pool.id === sharesOwned.poolId.id);
-
-            if (!pool) {
-                continue;
+        const poolIds = [];
```
danielmkm, line 95: Type your variables whenever possible, the `poolShares` above is not typed because the type is ensured in the return statement. I could push a boolean into your `poolIds` and it wouldn't complain

modules/portfolio/lib/portfolio-data.service.ts
```diff
@@ -92,35 +92,50 @@ export class PortfolioDataService {
...
-                ...pools.find((pool) => pool.id === sharesOwned.poolId.id)!,
-                poolId: pool.id,
-                amp: pool.amp ?? null,
-                blockNumber,
-                tokens: this.getPoolTokens(pool, pools).map((token) => ({
-                    ...token,
-                    snapshotId: '',
+        for (const pool of pools) {
+            // pool.address for masterchef subgraph, pool.id for sharesOwned in balancer subgraph
+            if (poolIds.includes(pool.address) || poolIds.includes(pool.id)) {
```
danielmkm, line 107: I see why you're doing this, but this is pretty confusing. At least rename your `poolIds` variable to something else. The pool Id is something different then the address

modules/portfolio/lib/portfolio-data.service.ts
franzns: Right, removed it.

modules/portfolio/lib/portfolio-data.service.ts
franzns: Gotcha. Done.

modules/portfolio/lib/portfolio-data.service.ts
franzns: You're right, it's stupid like this. Introduced a second array for the pool addresses. No need to save space. Better?

PR #27, opened 2022-05-14
Add tarot boosted apr

groninge01:
image

I added bignumber.js to make the conversion of the apr less funky.

### DIFF
--- a/modules/boosted/tarot.service.ts
+++ b/modules/boosted/tarot.service.ts
@@ -10,7 +10,7 @@ import { formatFixed } from '@ethersproject/bignumber';
 const TAROT_CACHE_KEY = 'tarot';
 
 interface TarotData {
-    address: string;
+    mainToken: string;
     apr: number;
 }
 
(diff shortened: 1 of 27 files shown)


danielmkm (changes requested):

package.json
```diff
@@ -26,6 +26,7 @@
         "apollo-server-core": "^3.5.0",
         "apollo-server-express": "^3.5.0",
         "axios": "^0.24.0",
+        "bignumber.js": "^9.0.2",
```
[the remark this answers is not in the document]
groninge01: @danielmkm I have removed bignumber.js and have used the ethers version. I also added xtinspirit & xtarot.

modules/boosted/tarot.service.ts
```diff
@@ -0,0 +1,105 @@
...
+}
+
+const TAROT_APR_CONTRACT_ADDRESS = '0xc32B9a98F2419544070E9580166C5A18Af8ADf9f';
+const tarotAPRContract = getContractAt(TAROT_APR_CONTRACT_ADDRESS, TarotBoostedAbi);
+
+export class TarotService {
+    private cache: CacheClass<string, TarotData[]>;
+    private boostedPools = [
+        {
+            address: '0xfB98B335551a418cD0737375a2ea0ded62Ea213b', // MAI
```
danielmkm, line 24: don't hardcode these addresses like this.
- Use the factory address on the pool to determine its a tarot pool
- Use the `mainIndex` and `wrappedIndex` on the pool to determine main/wrapped tokens.

danielmkm (changes requested):

modules/boosted/tarot.service.ts
groninge01: I have hardcoded the factory address (for now). Is there a way to get it from the balancer sdk config? Do I need to resolve the 'network' key somehow first to get to the 'linearPools' key?

groninge01: @danielmkm I figured out how to grab the factory address from the balancer sdk so it is no longer hardcoded anymore.

PR #55, opened 2022-08-08
double search window

### LINKED ISSUE #47: Handle case for missing block when querying for last day data
Sometimes when trying to find the block from 24h ago, there is no block found leading into a invalid property read. 

```ts
const subgraphPools = await this.balancerSubgraphService.getAllPools({ block: { number: parseInt(block24hAgo.number) } }, false);
```
Leads to

```
TypeError
Cannot read properties of undefined (reading 'number')
```

franzns:
increase search window size to find block 24h ago

resolves #47


0xSkly (approved):
Let's see if thats enough

Issue #63, opened 2022-08-09
my investments show pool with 0$ investment, bpt dust in wallet


franzns:
Address 0x500F8fe004E978D1c58dba852BfE7bBa8Ae434c1 shows pool https://v2.beets.fi/pool/0xdfc65c1f15ad3507754ef0fd4ba67060c108db7e000000000000000000000406 in my investment list although 0$ are invested. Have had investments in that pool long time ago.

![Image](https://user-images.githubusercontent.com/93920061/183618265-952aa6ee-df3f-406d-859e-ba0c189d3195.png)

0xSkly:
might get resolved by #64

franzns:
Oh, I think it's because I have BPT dust in my wallet.

0xSkly:
ah yea, we should filter them out in the backend

0xSkly:
closing in favour of https://github.com/beethovenxfi/beets-frontend/issues/123

PR #64, opened 2022-08-09
Update last synced block when no balance changes occurred

0xSkly:
If no balance changes occurred within the max block range, the function returns early without updating the latest synced block which is fixed by this PR


0xSkly:
resolves #63

Issue #65, opened 2022-08-09
Expose last synced block for wallet and staked balances in query


0xSkly:
We should expose the latest synced block for wallet & staked balances on a query.

I would propose to expose it on the protocol resolver.

The data can be found in the `PrismaUserBalanceSyncStatus` table.

![Image](https://user-images.githubusercontent.com/91403802/183650959-de49a134-0f2f-446b-9204-1eb31653300b.png)

danielmkm:
Would also be nice to have a mutation to force a balance update for a user address

0xSkly:
you mean like the `userSyncBalance` mutation just for all pools ?

franzns:
I'd also add the latest synced block for pool sync, needed for defilama.

So I propose to add the following data to the protocolMetrics:
```
    userWalletSyncBlock: Scalars['BigInt'];
    userStakeSyncBlock: Scalars['BigInt'];
    poolSyncBlock: Scalars['BigInt'];
```

franzns:
> Would also be nice to have a mutation to force a balance update for a user address

You'd like an update for all pools for a specific user address? check #71

franzns:
#71 merged

PR #69, opened 2022-08-10
Expose latest sync block

franzns:
addresses #65 

also adds black_list filter to gqlpoolsquery

### DIFF
--- a/modules/pool/pool.gql
+++ b/modules/pool/pool.gql
@@ -494,6 +494,7 @@ input GqlPoolFilter {
 
 enum GqlPoolFilterCategory {
     INCENTIVIZED
+    BLACK_LISTED
 }
 
 enum GqlPoolFilterType {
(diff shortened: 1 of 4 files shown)


modules/protocol/protocol.service.ts
```diff
@@ -50,13 +53,28 @@ export class ProtocolService {
...
+        });
+
+        const userWalletSyncBlock = await prisma.prismaUserBalanceSyncStatus.findUnique({
+            where: { type: PrismaUserBalanceType.WALLET },
+        });
+
+        const poolSyncBlock = await prisma.prismaLastBlockSynced.findUnique({
+            where: { category: PrismaLastBlockSyncedCategory.POOLS },
+        });
+
```
0xSkly, line 68: i would prefer this in its own query, since this is cached, so its not up to date and its queried by the frontend regularly which is not showing the data, so its triggering those db queries even though its not used.

Guess you should also be able to get the 3 in a single `findMany` query

modules/pool/pool.gql
```diff
@@ -494,6 +494,7 @@ input GqlPoolFilter {
 
 enum GqlPoolFilterCategory {
     INCENTIVIZED
+    BLACK_LISTED
```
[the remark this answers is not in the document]
franzns: yes

modules/protocol/protocol.service.ts
franzns: Refactored it into its own, non-cached query.

I like the approach with three simple DB calls better than looping and IFing over a result array. Not sure if you had something else in mind.

0xSkly (approved):
looks good i'd say

PR #80, opened 2022-08-15
Use contract events instead of changelog subgraph

### LINKED ISSUE #66: Replace changelog subgraph with on chain listener
We should try to get rid of the changelog subgraph and replace it with an on chain listener

0xSkly:
resolves #66


franzns (approved):
lgtm

Issue #61, opened 2022-08-08
Review cron job intervals


0xSkly:
We should investigate the cron job intervals and try to reduce them as much as possible which keeping the best user experience possible

franzns:
Not a lot of skips are happening so from that perspective no need to lower intervals. Any other reason why to reduce them? Otherwise I'd close this @0xSkly

PR #90, opened 2022-08-18
more meaningful log output

franzns:
wanted to take a look at #61 but need better output for that. Are you ok with the changes?


0xSkly (approved):
sure!

PR #101, opened 2022-08-23
Gql nullfield

franzns:
debug output

### DIFF
--- a/modules/pool/pool.service.ts
+++ b/modules/pool/pool.service.ts
@@ -108,10 +108,16 @@ export class PoolService {
 
         return batchSwaps.map((batchSwap) => ({
             ...batchSwap,
-            swaps: batchSwap.swaps.map((swap) => ({
-                ...swap,
-                pool: pools.find((pool) => pool.id === swap.poolId)!,
-            })),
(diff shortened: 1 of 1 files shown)


modules/pool/pool.service.ts
```diff
@@ -94,10 +94,16 @@ export class PoolService {
...
         return batchSwaps.map((batchSwap) => ({
             ...batchSwap,
-            swaps: batchSwap.swaps.map((swap) => ({
-                ...swap,
-                pool: pools.find((pool) => pool.id === swap.poolId)!,
-            })),
+            swaps: batchSwap.swaps.map((swap) => {
+                const pool = pools.find((pool) => pool.id === swap.poolId)!;
+                if (!pool) {
+                    console.log(`Missing pool for swap ${swap} in batchSwap ${batchSwap}`);
```
0xSkly, line 100: since `swap` is an object, i think `for swap ${swap}` would only log `[object object]`, so you need to either do a `JSON.stringify(swap)` or log an actually property

0xSkly (changes requested):

PR #118, opened 2022-09-01
Promote to production

franzns:
- remove unused changelog subgraph
- redo sync from sanity, also removes tokens from the whitelist
- token filters are not using AND
- minor fix in routes, called wrong method
- rename some sync methods to reflect what they are actually doing (only sync changes)

testing on https://canary.beets.fi/


0xSkly (approved):
Looks good i'd say

PR #119, opened 2022-09-05
get all snapshots for range

franzns:


modules/pool/pool.service.ts
```diff
@@ -124,6 +126,15 @@ export class PoolService {
         return featuredPoolGroups;
     }
 
+    public async getSnapshotsForAllPools(range: GqlPoolSnapshotDataRange) {
+        const allPools = await poolService.getGqlPools({});
```
0xSkly, line 130: so you do this madness query with all those joins just to get the poolId? also be aware that this filters out low liquidity pools. cant you just get the snapshots directly?

franzns: xD

PR #124, opened 2022-09-07
Handle empty rewarder

franzns:


modules/pool/lib/staking/fantom/master-chef-staking.service.ts
```diff
@@ -63,7 +66,11 @@ export class MasterChefStakingService implements PoolStakingService {
             if (farm.rewarder) {
                 for (const rewardToken of farm.rewarder.rewardTokens || []) {
                     const id = `${farmId}-${farm.rewarder.id}-${rewardToken.token}`;
-                    const rewardPerSecond = formatFixed(rewardToken.rewardPerSecond, rewardToken.decimals);
+                    const erc20Token = await getContractAt(rewardToken.token, ERC20Abi);
```
0xSkly, line 69: question if we should not batch all those calls in a multicall? but guess also ok like this

PR #148, opened 2022-09-16
use rewards per second from DB

franzns:
Take from DB as it already accounts for empty rewarder. Other option would be to have another RPC call, which seems more overhead.

### DIFF
--- a/modules/pool/lib/apr-data-sources/fantom/masterchef-farm-apr.service.ts
+++ b/modules/pool/lib/apr-data-sources/fantom/masterchef-farm-apr.service.ts
@@ -40,7 +40,7 @@ export class MasterchefFarmAprService implements PoolAprService {
             const totalLiquidity = pool.dynamicData?.totalLiquidity || 0;
             const farmTvl = totalShares > 0 ? (farmBptBalance / totalShares) * totalLiquidity : 0;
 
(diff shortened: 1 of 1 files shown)


modules/pool/lib/apr-data-sources/fantom/masterchef-farm-apr.service.ts
```diff
@@ -99,12 +99,16 @@ export class MasterchefFarmAprService implements PoolAprService {
...
         }
 
-        (farm.rewarder?.rewardTokens || [])
-            .filter((rewardToken) => rewardToken.token !== networkConfig.beets.address)
-            .forEach((rewardToken) => {
+        if (farm.rewarder) {
+            for (const rewardToken of farm.rewarder?.rewardTokens) {
+                const farmRewarder = await prisma.prismaPoolStakingMasterChefFarmRewarder.findUniqueOrThrow({
+                    where: {
+                        id: `${farm.id.toLowerCase()}-${farm.rewarder?.id.toLowerCase()}-${rewardToken.token.toLowerCase()}`,
```
0xSkly, line 106: that makes me cringe a bit. thats why it would have been better to use combined primary keys. but yea its what it is :)

0xSkly (approved):

PR #190, opened 2022-10-14
Fix fbeets balance sync

### LINKED ISSUE #181: fbeets are not correctly accounted for in pool balances
Fbeets show up in the userPoolBalance. Not 100% sure what the root cause is, could be either:
- all new fbeets holders experience that, e.g. only user that where already in the pool when the initBalance() happened work correctly
- when you get sent fbeets to a new wallet address, it's handled incorrectly

What I have seen:

Image

- User `0x5a9195392b95f49ac957deb7f991e9198138188b` got sent fbeets and staked them. They are also returned in poolBalances, which is incorrect. It is returned because it has set a pool ID in the DB:

Image
 
(poolId is `null` for users where it is working correctly)

0xSkly:
fbeets wallet and staking balance should have no poolId set. resolves #181

### DIFF
--- a/modules/user/lib/user-sync-wallet-balance.service.ts
+++ b/modules/user/lib/user-sync-wallet-balance.service.ts
@@ -1,19 +1,18 @@
-import { balancerSubgraphService } from '../../subgraphs/balancer-subgraph/balancer-subgraph.service';
-import { AddressZero } from '@ethersproject/constants';
-import { jsonRpcProvider } from '../../web3/contract';
-import ERC20Abi from '../abi/ERC20.json';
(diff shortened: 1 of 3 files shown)


modules/user/lib/user-sync-wallet-balance.service.ts
```diff
@@ -156,9 +155,14 @@ export class UserSyncWalletBalanceService {
                 ...balances
                     .filter(({ userAddress }) => userAddress !== AddressZero)
                     .map((userBalance) => {
+                        if (userBalance.erc20Address === networkConfig.fbeets.address) {
```
franzns, line 158: better use isSameAddress()

PR #196, opened 2022-10-25
Feature/datastudio feed

### LINKED ISSUE #121: work with souvlaki to support end of month reporting

franzns:
First try. Getting the private key from secrets has not yet been tested, need to test that on aws once deployed. When getting it from a file, the feed does was it should. Datasheet IDs might need to be updated, these are test sheets.

### DIFF
--- a/modules/datastudio/datastudio.service.ts
+++ b/modules/datastudio/datastudio.service.ts
@@ -141,19 +141,6 @@ export class DatastudioService {
             );
             allPoolDataRows.push(poolDataRow);
 
-            const poolTypeRow: string[] = [];
-            poolTypeRow.push(pool.address, pool.name, pool.id, pool.type, swapFee, pool.symbol);
-            allPoolTypeRows.push(poolTypeRow);
-
-            const allTokens = pool.allTokens.map((token) => {
-                const poolToken = pool.tokens.find((poolToken) => poolToken.address === token.token.address);
-
-                return {
-                    ...token.token,
-                    weight: poolToken?.dynamicData?.weight,
-                };
-            });
-
             for (const token of allTokens) {
                 const poolCompositionRow: string[] = [];
                 poolCompositionRow.push(
@@ -173,12 +160,6 @@ export class DatastudioService {
             networkConfig.datastudio[env.DEPLOYMENT_ENV as DeploymentEnv].databaseTabName,
             'A1:T1',
             allPoolDataRows,
-            jwtClient,
-        );
-        this.updateDataInSheet(
-            networkConfig.datastudio[env.DEPLOYMENT_ENV as DeploymentEnv].poolTypesTabName,
-            `A2:F${allPoolTypeRows.length + 1}`,
-            allPoolTypeRows,
             jwtClient,
         );
 
--- a/modules/datastudio/datastudio.service.ts
+++ b/modules/datastudio/datastudio.service.ts
@@ -1,0 +1,234 @@
+import { prisma } from '../../prisma/prisma-client';
+import { google } from 'googleapis';
+import { env } from '../../app/env';
+import { DeploymentEnv, networkConfig } from '../config/network-config';
+import moment from 'moment-timezone';
+import { JWT } from 'google-auth-library';
+import { SecretsManagerClient, GetSecretValueCommand } from '@aws-sdk/client-secrets-manager';
+
+export class DatastudioService {
+    public async feedPoolData() {
+        const secretManagerClient = new SecretsManagerClient({ region: env.AWS_REGION });
+
+        let privateKey = '';
+        try {
+            const response = await secretManagerClient.send(
+                new GetSecretValueCommand({ SecretId: 'backend-v2-datafeed-privatekey' }),
+            );
+            if (response.SecretString) {
+                privateKey = response.SecretString;
+                console.log(privateKey);
+            }
+        } catch (error) {
+            throw Error(`Error getting secret from AWS: ${error}`);
+        }
+
+        const jwtClient = new google.auth.JWT(
+            networkConfig.datastudio[env.DEPLOYMENT_ENV as DeploymentEnv].user,
+            undefined,
+            privateKey,
+            'https://www.googleapis.com/auth/spreadsheets',
+        );
+        await jwtClient.authorize(function (err, result) {
+            if (err) {
+                console.log(`Error authorizing: ${err}`);
+                return;
+            }
+        });
+
+        const sheets = google.sheets({ version: 'v4' });
+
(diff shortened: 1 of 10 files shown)


franzns:
The cron setup is a bit hacky. I need it to run as close to midnight as possible and all runs should be 24hours apart. That is the best approach I could come up with.

modules/datastudio/datastudio.service.ts
```diff
@@ -0,0 +1,215 @@
...
+import { google } from 'googleapis';
+import { env } from '../../app/env';
+import { DeploymentEnv, networkConfig } from '../config/network-config';
+import moment from 'moment-timezone';
+import { JWT } from 'google-auth-library';
+import { SecretsManagerClient, GetSecretValueCommand } from '@aws-sdk/client-secrets-manager';
+
+export class DatastudioService {
+    public async feedPoolData() {
+        const secretManagerClient = new SecretsManagerClient({ region: env.AWS_REGION });
```
0xSkly: i would pass this in the constructor, so it can be mocked out easier in tests

modules/datastudio/datastudio.service.ts
```diff
@@ -0,0 +1,215 @@
...
+
+        let privateKey = '';
+        try {
+            const response = await secretManagerClient.send(
+                new GetSecretValueCommand({ SecretId: 'backend-v2-datafeed-privatekey' }),
+            );
+            if (response.SecretString) {
+                privateKey = response.SecretString;
+                console.log(privateKey);
```
0xSkly: not sure you wanna log that private key :P

modules/datastudio/datastudio.service.ts
```diff
@@ -0,0 +1,215 @@
...
+            }
+        } catch (error) {
+            throw Error(`Error getting secret from AWS: ${error}`);
```
0xSkly: if you pass the error like this with string interpolation, it will internally call error.toString() which if the error is an object will just result in an unusable [object] output. Also why wrap the error again in another error? if you dont handle it and just rethrow it, just dont catch it. think the original error is probably more useful than your wrapped error

modules/datastudio/datastudio.service.ts
```diff
@@ -0,0 +1,215 @@
...
+        }
+
+        const jwtClient = new google.auth.JWT(
```
0xSkly: same as secrets manager, would pass it as constructor arg. you can do a `secrets-manager.ts` and `jwt- client.ts` file where those clients get created and exported, then you can easily pass them in the constructor

modules/datastudio/datastudio.service.ts
```diff
@@ -0,0 +1,215 @@
...
+        await jwtClient.authorize(function (err, result) {
+            if (err) {
+                console.log(`Error authorizing: ${err}`);
+                return;
+            }
+        });
+
+        const sheets = google.sheets({ version: 'v4' });
+
+        const range = `${networkConfig.datastudio[env.DEPLOYMENT_ENV as DeploymentEnv].databaseTabName}!B2:B`;
```
0xSkly: also configs i would pass in the constructor, if you wanna test this stuff you would need to mock out the whole network configs file

modules/datastudio/datastudio.service.ts
```diff
@@ -0,0 +1,215 @@
...
+        let result;
+        try {
+            result = await sheets.spreadsheets.values.get({
+                auth: jwtClient,
+                spreadsheetId: networkConfig.datastudio[env.DEPLOYMENT_ENV as DeploymentEnv].sheetId,
+                range: range,
+            });
+        } catch (e) {
+            throw Error(`Error reading values: ${e}`);
```
0xSkly: Same here, just cont catch the error if you dont handle it and rethrow it anyways, the `${e}` will probably not print anything useful if its not just a string

modules/datastudio/datastudio.service.ts
```diff
@@ -0,0 +1,215 @@
...
+        }
+
+        let lastRun = moment.tz('GMT').startOf('day').subtract(1, 'day').unix();
```
0xSkly, line 34: why gmt timezone and not UTC as we use utc for anything else?

modules/datastudio/datastudio.service.ts
```diff
@@ -0,0 +1,215 @@
...
+                    token.symbol,
+                    token.weight ? token.weight : `0`,
+                    pool.name,
+                );
+                allPoolCompositionRows.push(poolCompositionRow);
+            }
+        }
+
+        this.appendDataInSheet(
+            networkConfig.datastudio[env.DEPLOYMENT_ENV as DeploymentEnv].databaseTabName,
```
0xSkly: extract to constructor

modules/datastudio/datastudio.service.ts
```diff
@@ -0,0 +1,215 @@
...
+            'A1:T1',
+            allPoolDataRows,
+            jwtClient,
+        );
+
+        this.updateDataInSheet(
+            networkConfig.datastudio[env.DEPLOYMENT_ENV as DeploymentEnv].compositionTabName,
```
0xSkly: same

modules/datastudio/datastudio.service.ts
```diff
@@ -0,0 +1,215 @@
...
+                range: `${tabName}!${rowRange}`,
+                valueInputOption: 'USER_ENTERED',
+                requestBody: {
+                    majorDimension: 'ROWS',
+                    range: `${tabName}!${rowRange}`,
+                    values: rows,
+                },
+            });
+        } catch (e) {
+            throw Error(`Error updating values: ${e}`);
```
0xSkly: same here, would not catch it if you just rethrow it

modules/datastudio/datastudio.service.ts
```diff
@@ -0,0 +1,215 @@
...
+                range: `${tabName}!${rowRange}`,
+                valueInputOption: 'USER_ENTERED',
+                requestBody: {
+                    majorDimension: 'ROWS',
+                    range: `${tabName}!${rowRange}`,
+                    values: rows,
+                },
+            });
+        } catch (e) {
+            throw Error(`Error appending values: ${e}`);
```
0xSkly: same

0xSkly:
> The cron setup is a bit hacky. I need it to run as close to midnight as possible and all runs should be 24hours apart. That is the best approach I could come up with.

hm not sure what the approach is, you just run it every minute? 

edit: aha got it 
```ts
  if (lastRun > moment.tz('GMT').subtract(1, 'day').unix()) {
            // 24 hours did not pass since the last run
            return;
        }
```
mhh yea also not quite sure of a better way

modules/datastudio/datastudio.service.ts
franzns: same same :)

PR #213, opened 2022-11-14
support apr ranges

0xSkly:


franzns (approved):
Looks good. As daniel requested in the chat, let's add two new fields to the gql type: min and max

Question is, what to do with the total? I'd either set it to 0 or take max value.

0xSkly:
added the fields and using max in the totals

modules/pool/lib/pool-gql-loader.service.ts
```diff
@@ -458,18 +522,44 @@ export class PoolGqlLoaderService {
...
                 items: [
-                    ...aprItemsWithNoGroup.map((item) => ({
-                        ...item,
-                        apr: `${item.apr}`,
-                        subItems: [],
-                    })),
+                    ...aprItemsWithNoGroup.flatMap((item) => {
+                        let apr = `${item.apr}`;
+                        if (item.range) {
+                            apr = `${item.range.min} - ${item.range.max}`;
```
franzns: this still needed?

modules/pool/lib/pool-gql-loader.service.ts
```diff
@@ -417,12 +417,76 @@ export class PoolGqlLoaderService {
...
+                        maxThirdPartyApr += maxApr;
+                    }
+                    case 'SWAP_FEE': {
+                        swapFeeApr += maxApr;
+                        break;
+                    }
+                }
+            }
+            swapApr = `${swapFeeApr}`;
+            totalApr = `${maxTotalApr}`;
```
franzns, line 474: bullish

PR #239, opened 2023-01-19
add recoverymode flag for yieldfee

franzns:


modules/pool/lib/pool-on-chain-data.service.ts
```diff
@@ -152,7 +158,14 @@ export class PoolOnChainDataService {
 
             if (isComposableStablePool(pool) || isWeightedPoolV2(pool)) {
                 // the new ComposableStablePool and WeightedPool mint bpts for protocol fees which are included in the getActualSupply call
-                multiPool.call(`${pool.id}.totalSupply`, pool.address, 'getActualSupply');
+                // getActualSupply fails for not initialized pools (pools where totalSupply = 0)
+                // until it is initialized, we'll use the totalSupply since it is also also updated but not 100% accurate it should be good enough
+                if (pool.dynamicData && pool.dynamicData?.totalSharesNum === 0) {
+                    multiPool.call(`${pool.id}.totalSupply`, pool.address, 'totalSupply');
```
danielmkm, line 164: dunno if this is a good idea, just set it to 0

modules/pool/lib/pool-on-chain-data.service.ts
```diff
@@ -260,20 +273,26 @@ export class PoolOnChainDataService {
                     typeof onchainData.swapEnabled !== 'undefined'
                         ? onchainData.swapEnabled
                         : pool.dynamicData?.swapEnabled;
+                const recoveryModeEnabled =
+                    typeof onchainData.recoveryModeEnabled !== 'undefined'
+                        ? onchainData.recoveryModeEnabled
+                        : pool.dynamicData?.recoveryModeEnabled;
```
danielmkm, line 279: Better to be more distinct, just set it as false

modules/pool/lib/pool-on-chain-data.service.ts
```diff
@@ -290,7 +309,10 @@ export class PoolOnChainDataService {
                     const balance = formatFixed(poolTokens.balances[i], poolToken.token.decimals);
                     const weight = onchainData.weights ? formatFixed(onchainData.weights[i], 18) : null;
 
-                    let priceRate = onchainData.tokenRates ? formatFixed(onchainData.tokenRates[i], 18) : '1.0';
+                    let priceRate = '1.0';
+                    if (onchainData.tokenRates && onchainData.tokenRates[i]) {
+                        priceRate = formatFixed(onchainData.tokenRates[i], 18);
+                    }
 
                     if (onchainData.metaPriceRateCache && onchainData.metaPriceRateCache[i][0].gt('0')) {
```
danielmkm, line 317: should be `else if` here

danielmkm (commented):
small changes

danielmkm:
Don't see the case for when the pool is created

danielmkm:
Dont see the case for when the pool is created

modules/pool/lib/pool-on-chain-data.service.ts
danielmkm: scratch that, this will never get triggered then

franzns: when you set it to 0, then it will never update and use the actualSupply

PR #253, opened 2023-02-10
Support multiple chains in a single deployment

danielmkm:
This is a mega PR, but it's kinda an all or nothing, so let's rip the bandaid off and get it over with.

- Add `chain` column to all database tables (except `User`), convert primary keys to always include the chain id. So under most circumstances the primary key went from `id` -> `id + chain`. This ensures uniqueness across chains and allows for filtering for a specific chain across all tables.
- We were using a globally defined `networkConfig` constant that inferred the running network via an env var `env.CHAIN_ID`. This approach no longer works as incoming requests can be defined for any network (or no network for listing queries). Convert `networkConfig` to `networkContext`, a request scoped context service. This also means that accessing `networkContext` in the global scope is no longer supported, any reference to it needs to be inside a request scope, or more specifically, needs to always be referenced within a function call. For example we used a pattern of `new Service(networkConfig.address)` in the global scope quite often, this approach is no longer valid. The approach across the repo has been refactored to support the new request scoped model.
- Along the same lines as the previous comment, any usage of `env.CHAIN_ID` has been refactored to use `networkContext.chainId`.
- Request scoped context is implemented using `AsyncLocalStorage` following the simple pattern described here: https://medium.com/trabe/asynclocalstorage-for-easy-context-passing-in-node-js-e33c84679516. Renamed `userMiddleware` to `sessionMidleware` and extended it to support pulling the `ChainId` from the request header and injecting it into the `AsyncLocalStorage` session context. Inside of `NetworkContext`, we pull the active chainId from the session context, and use this to resolve network specific config.
- We isolate all network specific config into a single file in `network/${network}.ts` and register the model inside the `network/network-config.ts` file. The goal will be that adding support for a new network will only require the creation of this config file.
- For the worker, we currently expect that there will be 1 worker per network, but this can quite easily be changed. I will leave it to @franzns to weigh in on how he'd prefer this to be managed, and we can configure accordingly.

This is just a rough first pass and needs extensive testing. We'll also need to develop a migration strategy for fantom and optimism.


franzns:
Went through all the code and from a first glance looks good. Great work! 

I will create an AWS deployment next week and check if I run into any problems.

As for the workers, the plan is to use one worker for all chains and see how that performs with a configuration option to specify which workers should handle which chains. I will keep the worker environment setup on AWS EBS using SQS for now. This can be redone to a dockerized/k8s deployment at a later time.

franzns:
I have created a new branch `v3-canary` for this deployment.

PR #261, opened 2023-02-16
Make docker-compose launch all services and a setup script

timjrobinson:
Wanted to make it easy for anyone to spin this up locally with one command. 

- Extend docker-compose file to run the app, worker and setup script. 
- Move `scheduleWorkerTasks` to the worker file.
- Make worker run cron tasks on launch.
- Improve README, describing how to run the mutations and start the server.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,7 @@
+FROM node:16-slim
+
+WORKDIR /app
+
+RUN apt-get update -y && apt-get install -y openssl
+
+CMD ["yarn","start:local"]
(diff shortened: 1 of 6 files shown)


franzns:
Hey @timjrobinson thanks for the effort on this. A few things:
- Unfortunately this setup prevents me from running only the database in docker and the backend from the IDE for debugging purposes. Not sure how to achieve it though.
- I think the `scheduleLocalWorkerTasks()` doesn't include all necessary worker tasks, it wasn't maintained. Would need to compare the jobs yaml file with what this function runs.
- I'd like to introduce this only for v3, which is currently in the `feature/multiple-chains-single-deployment` branch. You mind creating a PR against that branch?

timjrobinson:
Closing for now, will re-work later on v3-main branch.

PR #266, opened 2023-02-20
Feature/sor query

franzns:
- Add queryBatchSwap call before returning SOR response
- Handle BAL#304 by trying to update liquidity once.

### DIFF
--- a/modules/balancer-sdk/balancer-sor.service.ts
+++ b/modules/balancer-sdk/balancer-sor.service.ts
@@ -13,9 +13,9 @@ import { AddressZero } from '@ethersproject/constants';
 import { Contract } from '@ethersproject/contracts';
 import VaultAbi from '../pool/abi/Vault.json';
 import { jsonRpcProvider } from '../web3/contract';
-import { balancerSdk } from './src/balancer-sdk';
 import { env } from '../../app/env';
-import { fp, fromFp, scaleDown } from '../big-number/big-number';
+import { Logger } from '@ethersproject/logger';
+import * as Sentry from '@sentry/node';
 
 interface GetSwapsInput {
     tokenIn: string;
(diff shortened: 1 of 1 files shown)


modules/balancer-sdk/balancer-sor.service.ts
```diff
@@ -46,67 +47,85 @@ export class BalancerSorService {
...
+            deltas = await this.queryBatchSwap(
+                swapType === 'EXACT_IN' ? SwapTypes.SwapExactIn : SwapTypes.SwapExactOut,
+                swapInfo.swaps,
+                swapInfo.tokenAddresses,
+            );
+        } catch (error: any) {
+            if (error.code === Logger.errors.CALL_EXCEPTION) {
+                const message: string = error.error.error.message;
+                // if we have shallow liquidity try to refresh pools and try again once
+                if (message.includes('BAL#304')) {
```
franzns: Is there a nice way to handle exceptions or is this what it looks like?

modules/balancer-sdk/balancer-sor.service.ts
```diff
@@ -46,67 +52,81 @@ export class BalancerSorService {
...
+        const tokenInAmountFixed = formatFixed(tokenInAmount, this.getTokenDecimals(tokenIn, tokens));
+        const tokenOutAmountFixed = formatFixed(tokenOutAmount, this.getTokenDecimals(tokenOut, tokens));
+
+        const effectivePrice = oldBnum(tokenInAmountFixed).div(tokenOutAmountFixed);
+        const effectivePriceReversed = oldBnum(tokenOutAmountFixed).div(tokenInAmountFixed);
         const priceImpact = effectivePrice.div(swapInfo.marketSp).minus(1);
 
+        for (const route of swapInfo.routes) {
+            route.tokenInAmount = oldBnum(tokenInAmountFixed).multipliedBy(route.share).toString();
+            route.tokenOutAmount = oldBnum(tokenOutAmountFixed).multipliedBy(route.share).toString();
```
danielmkm: on both of these, you're going to end up with more decimal places then you started with, breaking any calls to parseUnits later on. This needs adjusting.

danielmkm: You need to add a call to `dp` before toString()

oldBnum(tokenInAmountFixed).multipliedBy(route.share).dp(numDecimalsForThisToken).toString()

PR #285, opened 2023-03-17
CDK Setup Scripts

timjrobinson:
This adds CDK scripts to setup the BeethovenX Backend on AWS. 

They currently do the following:
- Setup a postgres database
- Setup a CodePipeline
- Setup 3 Elastic Beanstalk environments

All settings are based off the existing setup @franzns made.

There is probably a lot of small things to be fixed, and a lot needs to be cleaned up and de-duplicated, but the basics work. I'm not looking for a code review yet, just a functionality review, let me know about any differences between this stack and the existing setup that you see. 

Can be tested by switching to a different region than the existing setup and running `cdk bootstrap` (see https://docs.aws.amazon.com/cdk/v2/guide/bootstrapping.html) then `yarn deploy`.


buildspec.yml
```diff
@@ -18,7 +18,6 @@ phases:
         commands:
             - yarn generate
             - yarn prisma generate
-            - yarn prisma migrate deploy
```
timjrobinson, line 21: I think we should do the deploy during EBS init rather than during the build step (this is still TODO), it seems wrong for a build step to be modifying external services. What do you think?

PR #323, opened 2023-05-02
Set mainnet ETH address to 0xeee

timjrobinson:
The native asset address of Ethereum and Arbitrum were wrong. Setting these addresses to 0xeee so the price of the native token is in the token prices list for the Balancer frontend.


franzns:
Fixed this in a separate commit as this PR would also delete a prisma migration. Closing.

PR #332, opened 2023-05-08
remove the array on staking

danielmkm:


franzns:
@groninge01 this will break the v3 frontend again, but it's cleaner in the end like this.

PR #333, opened 2023-05-08
[WIP] Add b-sdk for trade queries

johngrantuk:
Initial work to complete Phase1 in this [Notion](https://www.notion.so/CowSwap-Solver-Update-4f75f0acb4824e2a8f293272b0e7b16e?pvs=4)

- Add new endpoints to query swap result (best of SORV1 vs V2):
  - sorGetSwaps - Beets (return formatted for existing Beets service)
  - sorGetCowSwaps - Balancer (return formatted for existing Balancer/CowSwap API)
- Add `@balancer/sdk` dependency for sorV2. (Needs updates for Fantom see below)
- Add Gyro pool fields required for swap maths (Note: Gyro pools must be deleted/refreshed to get new values from SG)

TODO:
- Gyro mainnet versioning is broken on SG which breaks SOR maths. A PR with fix is [here](https://github.com/balancer/balancer-subgraph-v2/pull/514)
- Gyro3 type needs updated once this PR is merged (https://github.com/balancer/b-sdk/pull/92), see TODO: modules/sor/sorV2/sorV2.service.ts
- Fantom:
  - b-sdk MulticallV3 update (to fix pool fetch issues) needs merged: https://github.com/balancer/b-sdk/pull/90
  - b-sdk PR to add chain needs merged: https://github.com/balancer/b-sdk/pull/69


modules/sor/sorV1/sorV1.service.ts
```diff
@@ -0,0 +1 @@
+// This will hold logic for querying SORV1 trade. Either via call to existing API or SORV1 package directly.
```
johngrantuk: Is it better to query the existing SOR API (which is SORV1) or bring that functionality into this backend? 
I guess there wouldn't be an issue with network delays so API call might be more efficient because the goal is to eventual remove SORV1 dependency.

modules/sor/sorV2/sorV2.service.ts
```diff
@@ -0,0 +1,243 @@
...
+export class SorV2Service {
+    public async getSwaps({
+        tokenIn,
+        tokenOut,
+        swapType,
+        swapAmount,
+    }: GetSwapsInput): Promise<GqlSorGetSwapsResponseNew> {
+        // TODO - This takes ~1.262s on my local machine. Is this fast enough? Any obvious ways to improve?
+        console.time('poolsFromDb');
+        const poolsFromDb = await this.getBasePoolsFromDb();
```
johngrantuk: Not sure how representative the timing from my local machine is compared to production?
Maybe I'm not approaching this correctly. Is there any obvious ways this can be improved?

modules/sor/sorV2/sorV2.service.ts
```diff
@@ -0,0 +1,243 @@
...
+     * @returns 
+     */
+    private mapToRawPools(pools: PrismaPoolWithDynamic[]): RawPool[] {
+        return pools.map(prismaPool => {
+            // b-sdk: src/data/types.ts
+            let rawPool: RawPool = {
+                id: prismaPool.id as Address,
+                address: prismaPool.address as Address,
+                poolType: this.mapRawPoolType(prismaPool.type),
+                poolTypeVersion: 1, // TODO - Can we add this to Prisma??
```
johngrantuk: Is it ok to add the pool type version to the database if its not already?

modules/sor/sorV2/sorV2.service.ts
```diff
@@ -0,0 +1,243 @@
...
+        // - LinearPoolType | 'Weighted' | 'Investment' | 'LiquidityBootstrapping' | 'Stable' | 'MetaStable' | 'ComposableStable' | 'StablePhantom' | 'Element';
+        switch(type) {
+            case PrismaPoolType.WEIGHTED:
+                return 'Weighted';
+            case PrismaPoolType.INVESTMENT:
+                return 'Investment';
+            case PrismaPoolType.LIQUIDITY_BOOTSTRAPPING:
+                return 'LiquidityBootstrapping';
+            case PrismaPoolType.STABLE:
+                return 'Stable'; // TODO - Is there a ComposableStable/Stable differentiation in Prisma?
```
johngrantuk: I couldn't see any differentiation between Stable/ComposableStable pool types. Is it possible to determine this somehow?

modules/sor/sor.service.ts
```diff
@@ -0,0 +1,60 @@
...
+            tokenOut,
+            swapType,
+            swapAmount,
+        });
+
+        let isSorV1 = false;
+        // TODO - Compare V1 vs V2 result and return/log best
+
+        // Update db with best result so we can track performace
+        await prisma.prismaTradeResult.create({
```
johngrantuk: Does this seem like an ok way to do this?

modules/sor/sorV1/sorV1.service.ts
franzns: Not sure what you mean by SOR API? For the Beethoven website we have the SOR deployed in a Lambda, is that what you mean?

modules/sor/sorV2/sorV2.service.ts
```diff
@@ -0,0 +1,243 @@
...
+                },
+            },
+            include: prismaPoolWithDynamic.include
+        });
+        const rawPools = this.mapToRawPools(pools);
+        // TODO - The pools below cause issue with b-sdk maths. Both have a token with balance = 0 so maybe that's issue?
+        const linearPoolsToIgnore = [
+            "0xbfa413a2ff0f20456d57b643746133f54bfe0cd20000000000000000000004c3", 
+            "0xdc063deafce952160ec112fa382ac206305657e60000000000000000000004c4", 
+        ]
```
franzns: These would be better configured in the network config. We also have some pools on FTM that we want to exclude from the SOR routing. Also it can be built into the prisma query.

modules/sor/sorV2/sorV2.service.ts
```diff
@@ -0,0 +1,243 @@
...
+import { networkContext } from '../../network/network-context.service';
+import { prisma } from '../../../prisma/prisma-client';
+import { PrismaPoolWithDynamic, prismaPoolWithDynamic } from '../../../prisma/prisma-types';
+import { RawPool } from '@balancer/sdk';
+import { HumanAmount, SupportedRawPoolTypes } from '@balancer/sdk';
+
+import { getSwapCompare } from './temp';
+
+// TODO - Check if this has been deployed to same address across networks?
+export const SOR_QUERIES = '0x1814a3b3e4362caf4eb54cd85b82d39bd7b34e41';
```
franzns: This should go into the network config

modules/sor/sorV2/sorV2.service.ts
franzns: This might not be much faster on the deployed instance. I'd probably use some caching here and only reload after the cache expired.

modules/sor/sorV2/sorV2.service.ts
franzns: Composablestables are phantomstables in prisma. There is an additional check isComposableStable() if you need the differentiation.

modules/sor/sorV2/sorV2.service.ts
```diff
@@ -0,0 +1,243 @@
+import { 
+    BasePool, 
+    ChainId,
```
franzns: can we get fantom added as a chainId in the SDK as well, please :)

modules/sor/sorV2/sorV2.service.ts
franzns: We don't track the version, but can be added of course. Question is where to source it from. I think balancer and beethoven use different versioning sometimes... Is that important for the SOR service?

modules/sor/sorV2/sorV2.service.ts
franzns: something similar to this https://github.com/beethovenxfi/beethovenx-backend/blob/sor-v2/modules/token/token.service.ts#L40-L50

modules/sor/sor.service.ts
franzns: yep looks good. only thing you could do is use composite keys instead of concated string keys, but that's taste in the end.

franzns: in the end, I'm not sure we want to save all the SOR queries into the DB. There are a shitload of queries sometimes. I think we can only persist a certain sample size.

modules/sor/sorV2/sorV2.service.ts
franzns: actually already is https://github.com/beethovenxfi/beethovenx-backend/blob/sor-v2/modules/network/network-config-types.ts#L87

modules/sor/sorV2/sorV2.service.ts
```diff
@@ -0,0 +1,243 @@
...
+        const chainId = networkContext.chainId as unknown as ChainId;
+        const tIn = await this.getToken(tokenIn as Address, chainId);
+        const tOut = await this.getToken(tokenOut as Address, chainId);
+        const swap = await sorGetSwapsWithPools(
+                    tIn,
+                    tOut,
+                    this.mapSwapType(swapType),
+                    swapAmount,
+                    poolsFromDb,
+                    // swapOptions, // TODO - Handle properly
```
franzns: We would also want to add default SOR swapOptions into the config, could be over written by passing to the resolver. Take a look at our lambda implementation https://github.com/beethovenxfi/sor-lambda/blob/main/src-ts/sorGetSwaps.ts#L51-L57

modules/sor/sorV2/sorV2.service.ts
johngrantuk: There's at least one case in new SOR where the maths differs depending on version.
I think having different versioning could make things more tricky. In the Balancer Subgraph there is the `poolTypeVersion` field. It is dependent on the pool factory.

modules/sor/sorV2/sorV2.service.ts
johngrantuk: I've moved to the `temp.ts` file for now as its only used there and I can hopefully remove this completely once the initial tests are done.

modules/sor/sorV2/sorV2.service.ts
johngrantuk: I'm hoping we can fix whatever is causing these specific pools to throw so we don't need to exclude them but for now I've moved them as you suggested.

modules/sor/sorV2/sorV2.service.ts
danielmkm: The versions for both beethoven and balancer are the same @franzns, the versions you're referring to are in the subgraph, but the one's @johngrantuk is referring to are on the factory/pool themselves

modules/sor/sor.service.ts
danielmkm: @johngrantuk's goal here is auditing past performance. I think this can be achieved by just logging them to cloudwatch. @franzns if you could provide an example of how to log to cloudwatch.

modules/sor/sorV2/sorV2.service.ts
danielmkm: Think it should be already

modules/sor/sor.prisma
```diff
@@ -0,0 +1,14 @@
+model PrismaTradeResult {
```
danielmkm: See other comment, would remove this model and just log this data to cloudwatch, its possible to run queries on the data.

modules/sor/sorV1/sorV1.service.ts
danielmkm: I would say use the API call, but you'll need to add an exception for networks that don't support the balancer API (fantom/optimism). In those instances, just do nothing.

modules/sor/sor.gql
```diff
@@ -0,0 +1,33 @@
+extend type Query {
+    sorGetSwapsNew(
```
danielmkm: You don't need to name this `New`, we should remove the old implementation.

modules/sor/sorV2/sorV2.service.ts
franzns: My bet is that these are uninitialized linears that fail on the getRate call. But the QueriesHelperContract should actually handle the failed call at least.

modules/sor/sorV1/sorV1.service.ts
franzns: or call our API :)

modules/sor/sorV2/sorV2.service.ts
franzns: I will add versioning from the subgraph into the DB.

modules/sor/sor.service.ts
franzns: @johngrantuk when you use console.log() it will log it to cloudwatch as well. In addition you could also publish a metric so we could see a nice dashboard how often v1 was better vs v2 and then go into the logs to investigate. For this, you would need to create another metrics client similar to this one https://github.com/beethovenxfi/beethovenx-backend/blob/sor-v2/modules/metrics/cron.metric.ts

modules/sor/sorV1/sorV1.service.ts
johngrantuk: Currently implemented using the Balancer V1 API. Will confirm with Tim what networks are currently supported and how hard it would be to get new ones added.

modules/sor/sor.service.ts
```diff
@@ -0,0 +1,82 @@
...
+            swapType,
+            swapAmount,
+        });
+        console.timeEnd('sorV2');
+
+        if(!swapV1.isValid && !swapV2.isValid) return EMPTY_COWSWAP_RESPONSE(tokenIn, tokenOut, swapAmount);
+
+        const bestSwap = this.getBestSwap(swapV1, swapV2, swapType, tokenIn, tokenOut);
+        
+        try {
```
franzns, line 39: Would remove this before merging

modules/sor/sor.service.ts
```diff
@@ -0,0 +1,82 @@
...
+import { EMPTY_COWSWAP_RESPONSE } from './constants';
+
+export class SorService {
+    public async getSwaps({
+        tokenIn,
+        tokenOut,
+        swapType,
+        swapAmount,
+    }: GetSwapsInput): Promise<GqlCowSwapApiResponse> {
+        console.time('sorV1');
```
franzns: should also add the networkcontext-chainId to the output

modules/sor/sor.service.ts
```diff
@@ -0,0 +1,82 @@
...
+
+        let isV1 = false;
+        if(!v1.isValid || !v2.isValid) {
+            isV1 = v1.isValid ? true : false;
+        }
+        else if(swapType === 'EXACT_IN') {
+            if(v2.outputAmount < v1.outputAmount) isV1 = true;
+        } else {
+            if(v2.inputAmount > v1.inputAmount) isV1 = true;      
+        }
```
franzns, line 108: should also log the chain

prisma/migrations/20230508110533_trade_results/migration.sql
```diff
@@ -0,0 +1,15 @@
...
+    "tokenIn" TEXT NOT NULL,
+    "tokenOut" TEXT NOT NULL,
+    "swapAmount" TEXT NOT NULL,
+    "swapType" TEXT NOT NULL,
+    "sorV1Result" TEXT NOT NULL,
+    "sorV2Result" TEXT NOT NULL,
+    "isSorV1" BOOLEAN NOT NULL,
+
+    CONSTRAINT "PrismaTradeResult_pkey" PRIMARY KEY ("id","chain")
+);
```
franzns: Simply remove both folders for the migrations

modules/sor/sor.resolvers.ts
```diff
@@ -0,0 +1,13 @@
+import { Resolvers } from '../../schema';
+import { sorService as sorService } from './sor.service';
+import { tokenService } from '../token/token.service';
+
+const sorResolvers: Resolvers = {
```
franzns: Since Beethoven and Balancer need to different return types for the sorGetSwaps, I'd suggest that we implement the resolvers under the Protocol specific resolvers in "modules/balancer" and "modules/beethoven"

franzns: From there we can trigger two different methods for the SOR v1 and adapt SOR v2 to serve both formats

modules/sor/sorV2/sorV2.service.ts
```diff
@@ -0,0 +1,332 @@
...
+                    totalSharesNum: {
+                        gt: 0.000000000001,
+                    },
+                    swapEnabled: true,
+                },
+                NOT: {
+                    id: {
+                        in: networkContext.data.sor[env.DEPLOYMENT_ENV as DeploymentEnv].poolIdsToExclude,
+                    }
+                }
```
franzns: this could be written as                 
id: {
        notIn: networkContext.data.sor[env.DEPLOYMENT_ENV as DeploymentEnv].poolIdsToExclude,
    },

modules/sor/sorV2/sorV2.service.ts
```diff
@@ -0,0 +1,332 @@
...
+        return this.mapToBasePools(rawPools);
+    }
+
+    /**
+     * Remove linear pools that cause issues because of price rate.
+     * @param pools 
+     */
+    private filterLinearPoolsWithZeroRate(pools: PrismaPoolWithDynamic[]): PrismaPoolWithDynamic[] {
+        // 0x3c640f0d3036ad85afa2d5a9e32be651657b874f00000000000000000000046b is a linear pool with priceRate = 0.0 for some tokens which causes issues with b-sdk
+        return pools.filter((p) => {
```
franzns: I'm a fan of verbose variable names, but I guess it comes down to taste :)

modules/sor/sorV2/sorV2.service.ts
```diff
@@ -0,0 +1,332 @@
...
+    /**
+     * Map Prisma pool type to b-sdk Raw pool type.
+     * @param type 
+     * @returns 
+     */
+    private mapRawPoolType(type: PrismaPoolType): SupportedRawPoolTypes | string {
+        // From b-sdk:
+        // - type LinearPoolType = `${string}Linear`;
+        // - LinearPoolType | 'Weighted' | 'Investment' | 'LiquidityBootstrapping' | 'Stable' | 'MetaStable' | 'ComposableStable' | 'StablePhantom' | 'Element';
+        switch(type) {
```
franzns: I guess this also needs gyro type?

modules/sor/sorV2/sorV2.service.ts
```diff
@@ -0,0 +1,332 @@
...
+        this.cache = new Cache<string, BasePool[]>();
+    }
+
+    public async getSwapResult({
+        tokenIn,
+        tokenOut,
+        swapType,
+        swapAmount,
+    }: GetSwapsInput): Promise<SwapResult> {
+        console.time('getBasePools');
```
franzns: if you want to leave this, add the chain ID

modules/sor/sorV2/sorV2.service.ts
```diff
@@ -0,0 +1,332 @@
...
+            return this.mapResultToCowSwap(this.swap, ip, op);
+        }
+    }
+
+    /**
+     * Maps Swap to GqlCowSwapApiResponse which is what current CowSwap Solver uses.
+     * @param swap 
+     * @returns 
+     */
+    private mapResultToCowSwap(swap: SwapSdk, inputAmount: TokenAmount, outputAmount: TokenAmount): GqlCowSwapApiResponse {
```
franzns: For Beethoven, it will need a 'mapResultToBeethovenSwap'

modules/sor/sorV1/sorV1.service.ts
```diff
@@ -0,0 +1,149 @@
...
+
+    private async querySorBeets(
+        swapType: string,
+        tokenIn: string,
+        tokenOut: string,
+        swapAmountScaled: string,
+        swapOptions: GqlSorSwapOptionsInput,
+    ) {
+        // Taken from: modules/beethoven/balancer-sor.service.ts
+        // TODO - Currently don't get a swap from mainnet. Need an example curl?
```
franzns: if this is only exposed on the Beethoven resolver it should be fine

modules/balancer/balancer.gql
```diff
@@ -1,7 +1,41 @@
+extend type Mutation {
+    balancerMutationTest: String!
+}
+
+union SorResponse = GqlSorGetSwapsResponse | GqlCowSwapApiResponse
+
 extend type Query {
     balancerQueryTest: String!
```
franzns: can be removed

modules/balancer/balancer.gql
```diff
@@ -1,7 +1,41 @@
+extend type Mutation {
+    balancerMutationTest: String!
+}
+
+union SorResponse = GqlSorGetSwapsResponse | GqlCowSwapApiResponse
```
franzns: this doesn't need to be a union, you can just return the cowswap response since this resolver is only visible in the balancer api

modules/pool/lib/pool-creator.service.ts
```diff
@@ -249,6 +249,21 @@ export class PoolCreatorService {
...
+                              lambda: pool.lambda || '',
+                              tauAlphaX: pool.tauAlphaX || '',
+                              tauAlphaY: pool.tauAlphaY || '',
+                              tauBetaX: pool.tauBetaX || '',
+                              tauBetaY: pool.tauBetaY || '',
+                              u: pool.u || '',
+                              v: pool.v || '',
+                              w: pool.w || '',
+                              z: pool.z || '',
+                              dSq: pool.dSq || '',
```
franzns, line 306: are these all fixed at creation time? no need to update?

modules/pool/pool.resolvers.ts
```diff
@@ -66,6 +66,9 @@ const balancerResolvers: Resolvers = {
         poolGetLinearPools: async () => {
             return poolService.getGqlLinearPools();
         },
+        poolGetGyroPools: async () => {
+            return poolService.getGqlGyroPools();
+        }
```
franzns, line 108: is this needed?

modules/pool/lib/pool-creator.service.ts
franzns: did you double check if we need all that data for the SOR and whether all gyro types have them?

modules/sor/sor.service.ts
```diff
@@ -0,0 +1,133 @@
...
+        try {
+            // Updates with latest onchain data before returning
+            return await bestSwap.getCowSwapResponse(true);
+        } catch (err) {
+            console.log(`Error Retrieving QuerySwap`);
+            console.log(err);
+            return EMPTY_COWSWAP_RESPONSE(tokenIn, tokenOut, swapAmount);
+        }
+    }
+
```
franzns, line 27: does this also need something for the balancer UI (in the future) like `getBalancerSwaps()`?

modules/sor/sorV1Beets/sorV1Beets.service.ts
```diff
@@ -0,0 +1,67 @@
...
+    async getCowSwapResponse(queryFirst = false): Promise<GqlCowSwapApiResponse> {
+        throw new Error('Use Balancer Service');
+    }
+
+    async getBeetsSwapResponse(queryFirst: boolean): Promise<GqlSorGetSwapsResponse> {
+        if (!this.isValid || this.swap === null) throw new Error('No Response - Invalid Swap');
+        // Beets service is already querying onchain
+        return this.swap;
+    }
+}
```
franzns, line 32: extract in its own file if possible

modules/beethoven/balancer-sor.service.ts
```diff
@@ -142,27 +115,70 @@ export class BalancerSorService {
...
+            marketSp: swapInfo.marketSp,
+            swaps: swapInfo.swaps,
+            tokenAddresses: swapInfo.tokenAddresses,
+            swapType,
+        });
+    }
 
-        const tokenInAmountFixed = formatFixed(tokenInAmount, this.getTokenDecimals(tokenIn, tokens));
-        const tokenOutAmountFixed = formatFixed(tokenOutAmount, this.getTokenDecimals(tokenOut, tokens));
+    formatResponse(swapData: {
```
franzns, line 136: I take it this is just a refactor?

modules/pool/lib/pool-creator.service.ts
johngrantuk: Fixed after pool creation. We need the data for the maths in SOR.

modules/pool/pool.resolvers.ts
johngrantuk: Is really useful for debugging.

modules/sor/sor.service.ts
johngrantuk: Not that I'm aware of.

modules/beethoven/balancer-sor.service.ts
johngrantuk: Yeah

gmbronco:
Both queries, sorGetSwaps and sorGetCowSwaps, will retrieve the optimal path from either V1 or V2.

Would you have any particular test cases in mind for testing these queries?

PR #350, opened 2023-05-23
Feat: ZKEVM Support

timjrobinson:
Added support for Polygon zkEVM chain.

Tested locally with https://github.com/balancer/frontend-v2/pull/3378 and it appears to be working

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,3 +87,7 @@ We follow the model of [gitflow](https://www.atlassian.com/git/tutorials/compari
 -   hotfix: hf/\*
 
 To contribute, branch from `v2-canary` (which is our development branch) and open a PR against `v2-canary` once the feature is complete. It will be reviewed and eventually merged into v2-canary.
+
+### Database Updates
+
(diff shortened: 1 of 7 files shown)


franzns:
As discussed, we are going to wait with this until the subgraphs are ready

timjrobinson:
This is good to go except for the smart contract deployed for `poolDataQueryContract`. 

We don't need the subgraphs yet as we're not going to have gauges for zkEVM on deployment and the APR's are being pulled from the Balancer API ATM (which doesn't need a blocks subgraph). I've removed the jobs related to staking.

franzns:
This won't work on other chains now as the ABI changed for the query helper contract. Can you adapt those configs as well? 

To be sure: You tested all the crons and none of them needs a blocks subgraph?

timjrobinson:
@franzns I've removed all crons that require the blocks subgraph and this is all working locally now with the new pool data contract. I've rebased it on #370 so that one can be merged first then this one.

PR #368, opened 2023-06-05
Default to swapEnabled=false when onchain sync fails.

johngrantuk:
There is an issue with swapEnabled state for some pools, for example: 
Mainnet: `0x3c640f0d3036ad85afa2d5a9e32be651657b874f00000000000000000000046b` is broken EulerLinear pool. It was previously ok so swapEnabled was set to true. 
Now when onchain state is trying to be sync'd the multicall throws an error and the pool sync gets skipped. 
This causes issues on b-sdk as the pool continues to show swapEnable=true so gets considered as a possible swap pool.

Not sure if what the effects are of setting other pools that throw during the multicall request but would assume if there's an issue with that then there is likely an issue with the pool. Alternative would be to add known broken pools to the ignore list so they don't get considered by b-sdk but this seems hard to maintain.

### DIFF
--- a/modules/pool/lib/pool-on-chain-data.service.ts
+++ b/modules/pool/lib/pool-on-chain-data.service.ts
@@ -65,7 +65,7 @@ const defaultPoolDataQueryConfig: PoolDataQueryConfig = {
 interface MulticallExecuteResult {
     targets?: string[];
     swapEnabled?: boolean;
-    pausedState?: [boolean, string, string]
+    pausedState?: [boolean, string, string];
 }
 
 const SUPPORTED_POOL_TYPES: PrismaPoolType[] = [
@@ -227,7 +227,13 @@ export class PoolOnChainDataService {
                 multiPool.call(`${pool.id}.swapEnabled`, pool.address, 'getSwapEnabled');
             }
 
-            if (pool.type === 'LINEAR' || pool.type === 'META_STABLE' || pool.type === 'PHANTOM_STABLE' || pool.type === 'STABLE' || pool.type === 'WEIGHTED') {
+            if (
+                pool.type === 'LINEAR' ||
+                pool.type === 'META_STABLE' ||
+                pool.type === 'PHANTOM_STABLE' ||
+                pool.type === 'STABLE' ||
+                pool.type === 'WEIGHTED'
+            ) {
                 multiPool.call(`${pool.id}.pausedState`, pool.address, 'getPausedState');
             }
         });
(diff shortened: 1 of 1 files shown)


modules/pool/lib/pool-on-chain-data.service.ts
```diff
@@ -310,12 +324,10 @@ export class PoolOnChainDataService {
                 const swapFee = formatFixed(poolData.swapFee, 18);
                 const totalShares = formatFixed(poolData.totalSupply, 18);
                 let swapEnabled: boolean | undefined;
-                if(typeof multicallResult?.swapEnabled !== 'undefined')
-                    swapEnabled =  multicallResult.swapEnabled;
-                else if(typeof multicallResult?.pausedState !== 'undefined')
+                if (typeof multicallResult?.swapEnabled !== 'undefined') swapEnabled = multicallResult.swapEnabled;
+                else if (typeof multicallResult?.pausedState !== 'undefined')
```
franzns: Can you please create proper IFs with curly brackets? :)

modules/pool/lib/pool-on-chain-data.service.ts
```diff
@@ -246,7 +252,15 @@ export class PoolOnChainDataService {
...
                 console.log(`Pool query return with error, skipping: ${poolData.id}`);
-                continue;
+                // Set swapEnabled to false to avoid issues with b-sdk considering these pools for swaps
+                await prisma.prismaPoolDynamicData.update({
+                    where: { id_chain: { id: poolData.id, chain: networkContext.chain } },
+                    data: {
+                        swapEnabled: false,
+                        blockNumber,
+                    },
+                });
```
franzns, line 262: This is true if the pool is "broken". I'm thinking about scenarios where a pool is being created, e.g. a linear pool that has just been created but not yet init joined. This would set this to false. I think it will set it to true once it updates because of the init join transaction. Can you double check if the init join triggers an update on the pool?

johngrantuk: Not sure I understand fully. I thought the onchain data was fetched on a cron job? So then it should be updated correctly when the pool is initialised because the pool query won't throw an error.

franzns: Only changed pools will be synced: https://github.com/beethovenxfi/beethovenx-backend/blob/swapEnabled-fix/modules/pool/lib/pool-sync.service.ts#L35-L37

But I assume an init join would also trigger a 'PoolBalanceChanged' event

franzns:
We just got a new version of the contract which now also offers a query for getPoolStatus to query recovery mode and paused state: https://ftmscan.com/address/0x7C47001Df19dF270A099a870474ee51c583C18F6#code

Maybe better to incorporate this contract? Might get rid of the case althogether

franzns:
gnosis: 0xC1Ff645400DD37989e77802326665cCf4fFDB352
fantom: 0x7C47001Df19dF270A099a870474ee51c583C18F6
optimism: 0x543a21679f2AAD98b04c6850D227812BaCe171E8
polygon: 0xb4761d0481B4f7a8A858D2796eEF3DAa2f3D9D2c
zkevm: 0x67af5D428d38C5176a286a2371Df691cDD914Fb8

mainnet pending

franzns:
we will be getting new contracts, never mind the addresses above!

franzns:
gnosis: 0x80dD2B80FbcFB06505A301d732322e987380EcD6
mainnet: 0x9Fa8A5977EF1d852602AE1960971D7ac0f0aaa1a
opera: 0xC1AFF8373840dB41275aAd5c41EC171D10872F5B
optimism: 0xd50C0e1abE53f706c41B400c587ddeBCadBF8045
polygon: 0x70b55Af71B29c5Ca7e67bD1995250364C4bE5554
zkevm: 0xC1Ff645400DD37989e77802326665cCf4fFDB352

PR #387, opened 2023-06-21
Feature/vebal voting list

agualis:
Adds logic to sync veBAL voting gauges and exposes and endpoint to get the list of veBAL voting pools.

The initial idea was automating the whole process and to fetch everything onchain but we found many different obstacles in the process. 
The most relevant ones:

- We cannot rely on the `network` field returned by the Gauge Controller contract because it is manually input by Maxis and it contains wrong networks.
- We cannot fully rely on the subgraph data because there are missing gauges (for instance, veBal, veLIT and veUSH).
- There are edge cases like some root gauge contracts missing fields (getRelativeWeightedCap) which involves multicalls allowing failures.

We still want to use fresher onchain data as much as possible in the API so the current solution is:

1) Pull the root gauge data onchain
2) Pull addresses, chains and recipient from the subgraph
3) Merge the data and save it in a new RootGauges table that will finally point to the related pool though different table relationships (RootGauge -> StakingGauge -> StakingPoolGauge -> Pool)
4) Expose a new endpoint that will return a list of pools with gauge info (relativeWeightedCap and gauge address)

There are some special cases that we handle with hardcoded data:
- veBal, veLIT, veUSH 
- Special pools (like TWAMM) that require hardcoding the whole response (pool and gauge data)


modules/network/arbitrum.ts
```diff
@@ -284,4 +284,40 @@ export const arbitrumNetworkConfig: NetworkConfig = {
             interval: every(5, 'minutes'),
         },
     ],
+    poolsWithGauges: [
```
agualis: This is a temporary solution to test the implementation until we get these pool lists in an automated way (probably pulling them from the RootGauges contract).

For the sake of testing, I just added this new configuration field to 2 network configs: MAINET and ARBITRUM

modules/network/mainnet.ts
```diff
@@ -183,10 +183,10 @@ export const mainnetNetworkConfig: NetworkConfig = {
     ],
     userStakedBalanceServices: [new UserSyncGaugeBalanceService()],
     /*
-    For sub-minute jobs we set the alarmEvaluationPeriod and alarmDatapointsToAlarm to 1 instead of the default 3. 
+    For sub-minute jobs we set the alarmEvaluationPeriod and alarmDatapointsToAlarm to 1 instead of the default 3.
```
agualis, line 183: This empty spaces are run by vscode formatter using the prettier settings from `.prettierrc` 

Given that there are no lint checks in CI, I guess it is ok to leave them but tell me if you have a concrete local setup that I should use.

modules/subgraphs/gauge-subgraph/gauge-subgraph-queries.graphql
```diff
@@ -155,3 +155,47 @@ query VotingEscrowLocks(
         }
     }
 }
+
+query GaugesInfo($poolIds: [Bytes!]) {
+    liquidityGauges(
+        where: { poolId_in: $poolIds, gauge_not: null }
+        # TODO: add pagination
```
agualis: We currently have less than 200 results in these queries. 

Will we ever reach 1000 results? We have 3 options here: 

1) Postpone the pagination decision (YAGNI?) by adding an exception when we detect for instance 900 results.  If that exception is thrown in the future it means that it is time to implement pagination
2) Implement pagination in GQL queries
3) Implement pagination in TS services

modules/subgraphs/gauge-subgraph/gauge-subgraph.service.ts
```diff
@@ -238,6 +244,29 @@ export class GaugeSubgraphService {
...
+    public async getGaugesStreamers(poolIds: string[]) {
+        const { liquidityGauges } = await this.sdk.GaugeStreamers({ poolIds });
+        return liquidityGauges;
+    }
+
+    public async getRootGaugesInfo(recipients: string[], chain: Chain): Promise<RootGaugesInfo> {
+        const currentChainId = getRequestScopeContextValue('chainId');
+        // Run query in MAINNET subgraph
+        // Can we do it in a cleaner way?
+        setRequestScopedContextValue('chainId', '1');
```
agualis: As far as I understand, the current architecture assumes that each job runs in a concrete network so it will setup the subgraph url depending on the current context network. However in this case (RootGauges), we need to query the MAINNET gauges subgraph independently of the current context network. Is there a better solution?

modules/vebal/voting-list.gql
```diff
@@ -0,0 +1,11 @@
+type VotingGauge {
+    address: String!
+    chain: String!
+    isKilled: Boolean!
+    createdAt: Date!
+    relativeWeightCap: String
+}
+
+extend type Query {
```
agualis: This is WIP, I still need to add the pool related data (I will ask some related questions soon).

modules/vebal/voting-list.service.ts
```diff
@@ -0,0 +1,188 @@
...
+            return pool;
+        });
+        // console.log('pools!', pools);
+        return Promise.resolve(pools);
+    }
+}
+
+export const votingListService = new VotingListService();
+
+async function fetchGaugesInfo(network: NetworkConfig): Promise<GaugesInfo> {
```
agualis: I added this functions outside the VotingListService() class so that it was easier to test them in isolation and to avoid `this.` noise. I can move them if that's the pattern that you always follow.

modules/vebal/voting-list.service.ts
```diff
@@ -0,0 +1,188 @@
...
+     * This query illustrates the pool related data that we need for each gauge
+     * Ideally, we don't want to denormalize but how can we join validGauges with getPoolsForVotingList in an efficient way?
+     */
+    public async getPoolsForVotingList(poolIds: string[]): Promise<any> {
+        let pools = await prisma.prismaPool.findMany({
+            where: { id: { in: poolIds } },
+            select: {
+                id: true,
+                address: true,
+                type: true, // TODO: Types are different than the original subgraph ones cause they are changed in mapSubgraphPoolTypeToPoolType (do we add a new column to prismaPool to store the original type?)
```
agualis: In frontend-v2 we use the original types from the subgraph. This is an issue because we need different LINEAR  types [that are lost after the mapping ](https://github.com/beethovenxfi/beethovenx-backend/blob/e25ef63fd7f88b4af72fe6291f19851d32220132/modules/pool/lib/pool-creator.service.ts#L413).
In the long term, we should probably try to find a common shared set of types for both frontends. 
In the short term, should we add a column "originalPoolType" or "balancerPoolType" in the tables using it?

modules/vebal/voting-list.service.ts
```diff
@@ -0,0 +1,188 @@
...
+    const calls = gaugeAddresses.map((address) => [
+        veBALHelperAddress,
+        false, // do not allow failures
+        veBalHelpers.encodeFunctionData('gauge_relative_weight', [address]),
+    ]);
+    // const result = await multicall.callStatic.aggregate3(calls);
+    // const weights = mapValues(result, weight => formatUnits(weight, 18));
+
+    //DEBUG: Return fake results until multicall is tested
+    return gaugeAddresses.reduce((acc: Record<string, string>, address: string) => {
```
agualis: For now, I use fake data to avoid using the real multicall. I will do that once we decide the overall design.

prisma/schema.prisma
```diff
@@ -821,4 +821,25 @@ model PrismaVeBalTotalSupply {
     chain               Chain
 
     totalSupply         String
-}
\ No newline at end of file
+}
+
+// QUESTION How do I add this automatically
+model PrismaVotingList {
```
agualis: Silly question: how do I automatically add the model from `voting-list.prisma` here?

worker/job-handlers.ts
```diff
@@ -256,6 +257,12 @@ export async function scheduleJob(job: WorkerJob, chainId: string) {
         case 'sync-vebal-totalSupply':
             await runIfNotAlreadyRunning(job.name, chainId, () => veBalService.syncVeBalTotalSupply(), 0.01);
             break;
+        case 'sync-vebal-voting-list':
+            //QUESTION: does it make sense to split it by chainId??
+            // TODO: define a proper sampling rate
+            const samplingRate = 0.01;
+            await runIfNotAlreadyRunning(job.name, chainId, () => votingListService.syncVotingList(), samplingRate);
```
agualis: The voting list is a feature that only works in MAINNET, however, in order to create the voting list we need to fetch info from all the L2 subgraphs. 
I designed it in a way that you only run it once and it writes the rows for all the networks. 
Following your typical job-handler flow,  we could easily split the single process in 1 process per network but, I see these trade-offs:

On one hand, having one job per network would make debugging a bit simpler: if the process fails in a concrete network we know which concrete network service to fix.
On the other hand, monitoring as much crons as networks can be extra cognitive/resource load that we don't really need.  

I'm sure that you have better arguments after your experience with the system, so happy to read them.

modules/vebal/voting-list.service.ts
```diff
@@ -0,0 +1,188 @@
...
+            })),
+        });
+    }
+
+    async getList(): Promise<PrismaVotingList[]> {
+        return await prisma.prismaVotingList.findMany();
+    }
+
+    /**
+     * This query illustrates the pool related data that we need for each gauge
```
agualis: This is one of the biggest pieces to finish the overall design of the feature: 

In the `veBalVotingList` resolver, I need to join the rows from PrismaVotingList with the pool data (from `prismaPool` and `prismaTokens`). 

`getPoolsForVotingList` is querying the pool data that I need (note that I also need to exclude BPT tokens from the result).

Each gauge row has a `poolId` column to perform the join with the previous data. I need a simple an efficient way to do that in the resolver but I never worked with prisma before, so I will need your help. 

Given that we number of rows is not huge, we could denormalise this pool data storing in the PrismaVotingList table. That would be super fast and flexible. But, I guess that you will have a better suggestion.

modules/vebal/voting-list.gql
franzns: Please define this asap so we can find out what data exactly is needed.

modules/subgraphs/gauge-subgraph/gauge-subgraph-queries.graphql
franzns: We implement pagingation in the TS service, see `getAllGaugeShares` for reference

modules/subgraphs/gauge-subgraph/gauge-subgraph.service.ts
franzns: I had to do something similar for the veBAL sync. I created a subgraph service called `vebal-locks-subgraph` which basically just points to the mainnet gauges subgraph. I'd suggest you rename that to `mainnet-gauges-subgraph` and also use it to get the root gauges.

modules/subgraphs/gauge-subgraph/gauge-subgraph.service.ts
```diff
@@ -238,6 +244,29 @@ export class GaugeSubgraphService {
...
 
+    public async getGaugesInfo(poolIds: string[]): Promise<LiquidityGaugesInfo> {
+        const { liquidityGauges } = await this.sdk.GaugesInfo({ poolIds });
+        return liquidityGauges;
+    }
+
+    public async getGaugesStreamers(poolIds: string[]) {
+        const { liquidityGauges } = await this.sdk.GaugeStreamers({ poolIds });
+        return liquidityGauges;
+    }
```
franzns: caution, v2 gauges dont use a streamer anymore

modules/vebal/voting-list.prisma
```diff
@@ -0,0 +1,19 @@
+model PrismaVotingList {
+    @@id([address, chain])
+
+    address             String
+    chain               Chain
+
+    isKilled            Boolean
+    createdAt           DateTime    @updatedAt
+    relativeWeightCap   String?
```
franzns: we already have the status of the gauges (killed, active, preferred) in the gauges table, no need to add it here. Only new field would be the weightCap as far as I understand? For just one field, we might add it to the gauges table instead of creatig a new table here

modules/vebal/voting-list.service.ts
franzns: Why do you need the different linear types? Why do you need the pooltypes at all?

modules/vebal/voting-list.service.ts
```diff
@@ -0,0 +1,188 @@
...
+     * Ideally, we don't want to denormalize but how can we join validGauges with getPoolsForVotingList in an efficient way?
+     */
+    public async getPoolsForVotingList(poolIds: string[]): Promise<any> {
```
franzns: No any types

modules/vebal/voting-list.service.ts
franzns: It's fine to do the join as you did below. Only thing I'm not sure is how we should handle gauges of pools which we don't have in the backend currently. There are these special type pools that are not supported but where there are hardcoded gauges https://github.com/balancer/frontend-v2/blob/master/public/data/hardcoded-gauges.json

modules/vebal/voting-list.service.ts
```diff
@@ -0,0 +1,188 @@
...
+            },
+        });
+
+        // Remove BPT
+        pools = pools.map((pool) => {
+            pool.tokens = pool.tokens.filter((t) => t.address !== pool?.address);
+            return pool;
+        });
+        // console.log('pools!', pools);
+        return Promise.resolve(pools);
```
franzns: why does this need another resolve?

modules/vebal/voting-list.service.ts
```diff
@@ -0,0 +1,188 @@
...
+    }
+}
+
+export const votingListService = new VotingListService();
+
+async function fetchGaugesInfo(network: NetworkConfig): Promise<GaugesInfo> {
+    const chainId = network.data.chain.id;
+    setRequestScopedContextValue('chainId', chainId.toString());
```
franzns: You don't need to set this here, it is sent in the query

modules/vebal/voting-list.gql
franzns: Also fine to just define it in json here in a comment and we can work from there. Might be easier as I think it's hard for you to determine the correct types here as you are not familiar with the types and data we have in other modules. I can then make a suggestion on the gql type returned here

modules/subgraphs/gauge-subgraph/gauge-subgraph.service.ts
franzns: But I would actually argue not to use the subgraph but rather the gauge controller contract https://etherscan.io/address/0xC128468b7Ce63eA702C1f104D55A2566b13D3ABD#readContract

modules/network/mainnet.ts
franzns: Strange, I also use the `.prettierrc` in the rootfolder of this project and it doesnt remove the space. Whatever, that is fine.

modules/vebal/voting-list.service.ts
agualis: We don't need them for this concrete feature, so let's forget this for now (probably YAGNI)

modules/vebal/voting-list.service.ts
```diff
@@ -0,0 +1,105 @@
...
+export const votingListService = new VotingListService();
+
+/**
+ * Fetches the relative weight of all gauges using the GaugeController Contract following this steps:
+ * 1) onchain call to know the total number of gauges
+ * 2) multicall (in chunks of 100) to get the list of gauge addresses
+ * 3) multicall (in chunks of 100) to get the gauge_relative_weight for each gauge address
+ */
+export async function fetchGaugeRelativeWeights(): Promise<Record<string, string>> {
+    const gaugeControllerAddress = '0xC128468b7Ce63eA702C1f104D55A2566b13D3ABD';
```
franzns: this would go into the network config

modules/vebal/voting-list.service.ts
```diff
@@ -0,0 +1,105 @@
...
+    const gaugeControllerContract = getContractAt(gaugeControllerAddress, gaugeControllerAbi);
+    const totalGauges = Number(formatFixed(await gaugeControllerContract.n_gauges()));
+
+    // There are two versions of gauge_relative_weight (argument override)
+    // so we filter the ABI to only include the one that we are using
+    const abi = gaugeControllerAbi.filter((item) => !(item.name === 'gauge_relative_weight' && item.inputs.length > 1));
```
franzns: this is not the same as the cap

modules/vebal/voting-list.service.ts
```diff
@@ -0,0 +1,105 @@
...
+
+export const votingListService = new VotingListService();
+
+/**
+ * Fetches the relative weight of all gauges using the GaugeController Contract following this steps:
+ * 1) onchain call to know the total number of gauges
+ * 2) multicall (in chunks of 100) to get the list of gauge addresses
+ * 3) multicall (in chunks of 100) to get the gauge_relative_weight for each gauge address
+ */
+export async function fetchGaugeRelativeWeights(): Promise<Record<string, string>> {
```
franzns: My proposal would be:

- Add a schema into prisma that holds the rootGauges with an ID, isKilled, relativeWeightCap and a relation to PrismaStakingGauge (will need to add an optional reference there too)
- For the sync the process would be:
1) get all gauge addresses like you did below here
2) for each gauge address you need to get the chain (controller.gauge_types(addr))
3) for each gauge you need to get `is_killed` and `getRelativeWeightCap`
4) if it's Mainnet, the gauge address will be the liquidity gauge of the pool -> associate to the PrismaStakingGauge (of not found, special manual pool like veBAL)
5) if it's an L2, you need to query `getRecipient` as well to find the liquidity gauge address -> associate to the PrismaStakingGauge (of not found, special manual pool like veBAL)

Now the special manual gauges I'm not yet sure how to handle... Probably we need to keep track of all gauges that we couldn't associate and make sure that we have them as manual entries, otherwise we need to fire an alarm or an exception. 

Don't yet know how to best integrate the manual gauges.

modules/vebal/voting-list.service.ts
agualis: Yeah, because we need this `relative_weight` to know when to exclude a killed gauge from the voting list.

modules/vebal/voting-list.service.ts
agualis: We only have one for all networks, so how would you add it? Optional field or different approach?

modules/vebal/voting-list.service.ts
agualis: Why storing isKilled again if we already have it in the linked `PrismaStakingGauge`?

The important thing here is that we want to discard gauges which: 
isKilled = TRUE
relativeWeight = 0.0

We can include a new column to show that concept (killed and empty) or we can simply avoid those gauges from the `rootGauges` table.

I would go for the second option and rename `rootGauges` to `votingGauges` so that: 

The FE runs a query that joins pools with votingGauges to the list of pools where we can vote (also linked to the PrismaStakingGauge info for each pool). Does it make sense?

modules/vebal/voting-list.service.ts
franzns: Optional field and only add for mainnet

modules/vebal/voting-list.service.ts
franzns: there is a `is_killed()` query on the gauge controller, shouldn't we use that one?

modules/vebal/voting-list.service.ts
franzns: > Why storing isKilled again if we already have it in the linked `PrismaStakingGauge`?

I'm not sure about the process, if I kill a root gauge, is the child gauge automatically killed as well? And vice versa?

> The important thing here is that we want to discard gauges which: isKilled = TRUE relativeWeight = 0.0

Can you elaborate what `relativeWeight` is and why you need to check for that in addition to `isKilled`?

> I would go for the second option and rename `rootGauges` to `votingGauges` so that:

I would stick to the rootGauge name, as this is exactly what they are called in the contracts. Also, we would add all rootGauges into the DB, not just the ones that we can vote on.

modules/network/viem/clients.ts
```diff
@@ -0,0 +1,12 @@
+import { createPublicClient, http } from 'viem';
+import { mainnet } from 'viem/chains';
+
+export function createHttpClient(httpRpc: string) {
+    return createPublicClient({
+        batch: {
+            multicall: { batchSize: 500 },
+        },
+        chain: mainnet,
```
franzns: remove this line

prisma/schema.prisma
```diff
@@ -875,4 +876,24 @@ model PrismaVeBalTotalSupply {
...
+  id      String
+  chain   Chain
+
+  staking PrismaPoolStakingGauge? @relation(fields: [stakingId, chain], references: [id, chain])
+  status              PrismaRootStakingGaugeStatus @default(ACTIVE)
+  gaugeAddress      String
+  stakingId         String?
+  relativeWeight    String?
+  relativeWeightCap String?
+}
```
franzns: this needs to be in it's own file in the module. it is merged here by using `yarn prisma-merge`

modules/vebal/root-gauges.onchain.ts
```diff
@@ -0,0 +1,221 @@
...
+    if (!Object.keys(Chain).includes(network)) throw Error(`Network ${network} is not supported`);
+    return network as Chain;
+}
+
+function generateGaugeIndexes(totalGauges: number) {
+    return [...Array(totalGauges)].map((_, index) => index);
+}
+
+export const veGauges = [
+    // They are not listed in the subgraph but we do have pool and staking info stored
```
franzns: They are not in the subgraph?

franzns: Ah, these are the recipients, not the rootGauges?

modules/vebal/root-gauges.onchain.ts
```diff
@@ -0,0 +1,221 @@
...
+        */
+        const gaugeTypeIndexes = await this.multicallGaugeController(gaugeAddresses, 'gauge_types');
+        const gaugeTypes = mapValues(gaugeTypeIndexes, (type) => typeNames[Number(type)]);
+
+        const isKilled = await this.multicallRootGauges(gaugeAddresses, 'is_killed');
+
+        const relativeWeightCapsWithFailures = await this.multicallRootGaugesAllowingFailures(
+            gaugeAddresses,
+            'getRelativeWeightCap',
+        );
```
franzns: are there rootGauges that don't have this field? Or why do we need to allow failures?

modules/vebal/root-gauges.onchain.ts
```diff
@@ -0,0 +1,221 @@
...
+        const relativeWeightCaps = mapValues(relativeWeightCapsWithFailures, (r) => r.result);
+
+        let rootGauges: RootGauge[] = [];
+        gaugeAddresses.forEach((gaugeAddress) => {
+            const relativeWeight = relativeWeightCaps[gaugeAddress];
+            if (gaugeTypes[gaugeAddress] === 'Liquidity Mining Committee') return;
```
franzns: What is this?

modules/vebal/root-gauges.prisma.ts
```diff
@@ -0,0 +1,76 @@
+import { Chain, PrismaClient } from '@prisma/client';
+import { Address } from 'viem';
+import { prisma as prismaClient } from '../../prisma/prisma-client';
+import { RootGauge, isValidForVotingList } from './root-gauges.onchain';
+
+export class PrismaRootGauges {
+    constructor(private prisma: PrismaClient = prismaClient) {}
```
franzns: can just use the exported prisma, no need for `this.prisma` below

modules/vebal/root-gauges.subgraph.ts
```diff
@@ -0,0 +1,55 @@
...
+    return [...l2RootGauges, ...mainnetRootGauges];
+}
+
+export function updateOnchainGaugesWithSubgraphData(onchainGauges: RootGauge[], subgraphGauges: SubGraphRootGauge[]) {
+    const subgraphGaugesByAddress = keyBy(subgraphGauges, 'gaugeAddress');
+
+    return onchainGauges.map((gauge) => {
+        const rootGauge = gauge;
+        const subGraphGauge = subgraphGaugesByAddress[gauge.gaugeAddress];
+        if (subGraphGauge) {
```
franzns: which root gauges are on the contract but not in the subgraph?

agualis: We have the following ones currently: 

        '0x7aa5475b2ea29a9f4a1b9cf1cb72512d1b4ab75e', //Liquidity Mining Committee gauge
        '0x56124eb16441a1ef12a4ccaeabdd3421281b795a', // veLIT
        '0x5b79494824bc256cd663648ee1aad251b32693a9', // veUSH
        '0xb78543e00712c3abba10d0852f6e38fde2aaba4d', // veBAL
        '0x3f829a8303455cb36b7bcf3d1bdc18d5f6946aea', // Arbitrum killed gauge that was replaced by 0xd758454bdf4df7ad85f7538dc9742648ef8e6d0a
        '0xe867ad0a48e8f815dc0cda2cdb275e0f163a480b', // killed
        '0x9fb8312cedfb9b35364ff06311b429a2f4cdf422', // killed

That's why I added a check to throw in case de have a missing alive one in the future

modules/vebal/root-gauges.prisma.ts
agualis: I used dependency injection so that injecting a in unit tests prima mock is easy: https://github.com/beethovenxfi/beethovenx-backend/blob/4eca19b736c7c69232e822aba9f1b27c073bfc91/modules/vebal/root-gauges.prisma.spec.ts#L22

Directly using the import would make tests more difficult (enforcing less direct and TS safe patterns like jest.mock).

modules/vebal/root-gauges.onchain.ts
agualis: It is an special root gauge: https://etherscan.io/address/0x7aa5475b2ea29a9f4a1b9cf1cb72512d1b4ab75e#readContract 

with type 0 (Liquidity Mining Committee):

LiquidityMinningComitee

We just exclude it.

modules/vebal/root-gauges.onchain.ts
agualis: Currently 126 out of 333 root gauge deployed contracts do not have this field. Examples: 

```
  '0xf0ea3559cf098455921d74173da83ff2f6979495',
  '0x3fdb6fb126521a28f06893f9629da12f7b7266eb',
  '0xe0b50b0635b90f7021d2618f76ab9a31b92d0094',
  '0x785f08fb77ec934c01736e30546f87b4daccbe50',
  '0x981fb05b738e981ac532a99e77170ecb4bc27aef',
  '0xb0de49429fbb80c635432bbad0b3965b28560177',
  '0x359ea8618c405023fc4b98dab1b01f373792a126',
  '0x435272180a4125f3b47c92826f482fc6cc165958',
  '0x05e7732bf9ae5592e6aa05afe8cd80f7ab0a7bea',
  '0xc77e5645dbe48d54afc06655e39d3fe17eb76c1c',
  '0xf30db0ca4605e5115df91b56bd299564dca02666',
  '0x899f737750db562b88c1e412ee1902980d3a4844',
  '0x6cb1a77ab2e54d4560fda893e9c738ad770da0b0',
  '0x6823dca6d70061f2ae2aaa21661795a2294812bf',
  '0xa5a0b6598b90d214eaf4d7a6b72d5a89c3b9a72c',
  '0x88d07558470484c03d3bb44c3ecc36cafcf43253',
  '0xfbf87d2c22d1d298298ab5b0ec957583a2731d15',
```

modules/vebal/root-gauges.onchain.ts
agualis: No, these are the root gauge addresses and they are not in the subgraph. Example of veBAL: 
address: 0xf2ca6f8961e91f1ee0d688f9926183314d866f1e

Let's talk to see how we can deal with these cases.

prisma/schema.prisma
agualis: Updated.

PR #390, opened 2023-06-23
Gyro dynamic data

johngrantuk:
- Previously thought GyroE V2 rates could be updated using scaling factor, this was incorrect
- Changed to query using `getTokenRates` via multicall
- Added new GyroDynamicData which adds tokenRates (this is needed by SOR)


modules/pool/lib/pool-on-chain-data.service.ts
```diff
@@ -402,6 +408,33 @@ export class PoolOnChainDataService {
...
+                                poolId: pool.id,
+                                tokenRates,
+                                blockNumber,
+                            },
+                            update: { tokenRates, blockNumber },
+                        });
+                    }
+                }
+
                 for (const poolToken of pool.tokens) {
```
franzns, line 438: You should add the rate under the poolToken instead of creating a new dataset for this. Similar to what we do for the linear token rates.

johngrantuk:
Replaced by #392

PR #396, opened 2023-07-03
APR Services - Add some yield tokens APR to linear pools

lgahdl:
Adding APR Service to fetch token yields APR on https://yield-tokens.balancer.workers.dev/ and calculate the APR for boosted pools with nested pools that contain the fetched tokens.


gmbronco:
What are "Non Reaper Boosted pools"?

lgahdl:
> What are "Non Reaper Boosted pools"?

I thought it would be necessary to update boosted pools in the same APR Service that updates the linear pools with "non reaper tokens" yield (Which John called Non Reaper Boosted Pools, because there's another service that updates the APR for pools with wrapped tokens from Reaper Crypt) but there's already a service that updates all boosted pools, so it's not necessary to update the "Non Reaper Boosted Pools", because they're already being updated by the "boosted-pool.apr.service"

modules/pool/lib/apr-data-sources/ib-tokens-apr.service.ts
```diff
@@ -0,0 +1,52 @@
...
+          const itemId = `${ pool.id }-tokens-yield-apr`
+          operations.push(prisma.prismaPoolAprItem.upsert({
+            where: { id_chain: { id: itemId, chain: networkContext.chain } }, create: {
+              id: itemId,
+              chain: networkContext.chain,
+              poolId: pool.id,
+              title: `${ token.token.symbol || token.address } APR`,
+              apr: aprs.get(token.address) ?? 0,
+              group: null,
+              type: 'IB_YIELD',
```
franzns: This needs to be different, based on the pool. If it's a wrapped asset from a linear pool, it needs to be `LINEAR_BOOSTED`, else `IB_YIELD`

modules/pool/lib/apr-data-sources/ib-tokens-apr.service.ts
```diff
@@ -0,0 +1,52 @@
...
+    await prismaBulkExecuteOperations(operations);
+  }
+
+  private async fetchYieldTokensApr(): Promise<Map<string, number>> {
+    const res = await axios.get("https://yield-tokens.balancer.workers.dev/");
+    const aprs = new Map<string, number>(Object.entries(res.data as {
+      [key: string]: number
+    }).map(([key, apr]) => [key, apr * 0.0001]));
+    return aprs;
+  }
```
franzns: I assume this will change to be sourced locally from the sources you added?

modules/token/lib/token-apr-handler/fetch-all.ts
```diff
@@ -0,0 +1,16 @@
+import { tokens } from "./tokens";
+import { TokenApr } from "./types";
+
+export const fetchAllAprs = async (): Promise<TokenApr[]> => {
```
franzns: Please create a proper function for this

modules/token/lib/token-apr-handler/sources/reaper.ts
```diff
@@ -0,0 +1,50 @@
+import { abi } from './abis/reaperStrategy'
+import { JsonRpcProvider } from "@ethersproject/providers";
+import { MulticallWrapper } from "ethers-multicall-provider";
+import { Contract } from "ethers";
+
+
+const jsonRpcProvider = new JsonRpcProvider('https://arb1.arbitrum.io/rpc', 42161 /*ARBITRUM*/)
+const provider = MulticallWrapper.wrap(jsonRpcProvider)
```
franzns: why is this multicall wrapper needed?

modules/token/lib/token-apr-handler/fetch-all.ts
```diff
@@ -0,0 +1,16 @@
...
+  const res = await Promise.all(
```
franzns: any of these fetches can fail for various reasons. I'd rather do that sequential (no need to optimize in the cron) and log proper errors to sentry

modules/token/lib/token-apr-handler/fetch-all.ts
```diff
@@ -0,0 +1,16 @@
...
+    tokens.map(async ({ name, group, fetchFn }) => {
+      const fetchedResponse: {[key:string]:number} = await fetchFn()
+      return Object.entries(fetchedResponse).map(([address, aprValue]) => ({
+        val: aprValue * 0.0001 /*Values come in BPS 10000=>100% */,
```
franzns: maybe the apr services should already provide values in %

franzns:
Looked through the thing now. We will need to redesign it a bit and make it multichain aware. Currently, some APR sources are multichain aware and others arent. 
I would even redesign the whole thing, introducing interfaces for the APR sources, similar to what we did for the other APR services. 
In addition, that thing should live in the pools module with the other APR code and not in the token module.

modules/token/lib/token-apr-handler/sources/reaper.ts
lgahdl: I used it so calls performed simultaneously can be automatically batched. Like in this case:
```
const getAprs = () => {

  const calls = Object.keys(strategiesMap).map(async (coin) => {
    const contract = new Contract(strategiesMap[coin], abi, provider)
    return contract.averageAPRAcrossLastNHarvests(3)
  })

  return Promise.all(calls)
}
```

on viem, it was this:
```
const getAprs = () => {
  const contracts = Object.keys(strategiesMap).map((coin) => ({
    address: strategiesMap[coin],
    abi,
    functionName: 'averageAPRAcrossLastNHarvests',
    args: [3],
  }))

  return client.multicall({ contracts })
}

```

modules/pool/lib/apr-data-sources/ib-yield-apr-handlers/ib-yield-apr-handlers.ts
```diff
@@ -0,0 +1,28 @@
...
+        address
+      }))
+    });
+    const res = Array(this.handlers.length)
+    for (const [index, aprPromise] of aprPromises.entries()) {
+      res[index] = await aprPromise
+    }
+    return res.flat();
+  }
+}
```
franzns: Make this a proper function of this class

modules/pool/lib/apr-data-sources/ib-yield-apr-handlers/sources/index.ts
```diff
@@ -0,0 +1,50 @@
...
+  eulerMainnetAprHandler,
+  gearboxMainnetAprHandler,
+  idleMainnetAprHandler,
+  overnightMainnetAprHandler,
+  ovixZkEVMAprHandler,
+  reaperArbitrumAprHandler,
+  tesseraApePoolAprHandler,
+  tetuAprHandler,
+  tranchessMainnetAprHandler,
+]
```
franzns: Should these better be initialized in the ib-yield-apr-handlers.ts instead of an additional file?

modules/pool/lib/apr-data-sources/ib-yield-apr-handlers/types.ts
```diff
@@ -0,0 +1,10 @@
+export type TokenApr = {
+  val: number;
+  address: string;
+  group?: string;
+}
+
+export interface AprHandler {
+  readonly group: string | undefined;
```
franzns: Should this ever be undefined?

modules/pool/lib/apr-data-sources/ib-tokens-apr.service.ts
```diff
@@ -0,0 +1,68 @@
...
+          operations.push(prisma.prismaPoolAprItem.upsert({
+            where: { id_chain: { id: itemId, chain: networkContext.chain } },
+            create: {
+              id: itemId,
+              chain: networkContext.chain,
+              poolId: pool.id,
+              title: `${ tokenSymbol } APR`,
+              apr: aprs.get(token.address)?.val ?? 0,
+              group: (aprs.get(token.address)?.group as PrismaPoolAprItemGroup) ?? null,
+              type: pool.type === 'LINEAR' ? 'LINEAR_BOOSTED' : 'IB_YIELD',
```
franzns: that is not enough, you'll need to check if it is the wrapped token when it is in a linear pool. only then it is linear_boosted, otherwise its IB_yield. E.g. for a wsteth/rfWSTETH linear pool

modules/pool/lib/apr-data-sources/ib-yield-apr-handlers/sources/index.ts
```diff
@@ -0,0 +1,25 @@
...
+  ...defaultHandlers,
+  ...eulerHandlers,
+  ...gearboxHandlers,
+  ...idleAprHandlers,
+  ...overnightHandlers,
+  ...ovixHandlers,
+  ...reaperHandlers,
+  ...tesseraHandlers,
+  ...tetuHandlers,
+]
```
franzns: include this into the ib-yield-apr-handlers.ts file, no need to do it separately

modules/pool/lib/apr-data-sources/ib-yield-apr-handlers/ib-yield-apr-handlers.ts
```diff
@@ -0,0 +1,26 @@
...
```
franzns: couldn't we do this is a simple for loop? this is super hard to read

modules/pool/lib/apr-data-sources/ib-yield-apr-handlers/ib-yield-apr-handlers.ts
```diff
@@ -0,0 +1,26 @@
...
+import { aprHandlers } from "./sources";
+
+export class IbYieldAprHandlers {
+  
+  private handlers: AprHandler[] = [];
+  constructor(network: number) {
+    this.handlers = aprHandlers.filter((handler) => handler.network === network);
+  }
+  
+   async getHandlersAprs (): Promise<TokenApr[]>  {
```
franzns: please rename to something like "fetchAprsFromAllHandlers"

modules/pool/lib/apr-data-sources/ib-yield-apr-handlers/types.ts
```diff
@@ -0,0 +1,10 @@
+export type TokenApr = {
+  val: number;
+  address: string;
+  group?: string;
+}
+
+export interface AprHandler {
+  readonly group: string;
+  getAprs(): Promise<{ [key: string]: number }>;
+}
```
franzns: could these types live in the ib-yield-apr-handlers.ts without creating circular dependencies?

modules/pool/lib/apr-data-sources/ib-yield-apr-handlers/types.ts
```diff
@@ -0,0 +1,10 @@
+export type TokenApr = {
+  val: number;
+  address: string;
+  group?: string;
+}
+
+export interface AprHandler {
+  readonly group: string;
+  getAprs(): Promise<{ [key: string]: number }>;
```
franzns: key is actually token address, right? rename it if so

modules/network/arbitrum.ts
```diff
@@ -160,6 +162,7 @@ export const arbitrumNetworkConfig: NetworkConfig = {
     contentService: new GithubContentService(),
     provider: new ethers.providers.JsonRpcProvider(arbitrumNetworkData.rpcUrl),
     poolAprServices: [
+        new IbTokensAprService(arbitrumNetworkData.chain.id),
         new WstethAprService(tokenService, arbitrumNetworkData.lido!.wstEthContract),
         new ReaperCryptAprService(
```
franzns: we should be able to remove these if we have the IbTokensAprService

modules/pool/lib/apr-data-sources/ib-yield-apr-handlers/sources/aave-apr-handler.ts
```diff
@@ -0,0 +1,232 @@
...
+            );
+            return aprEntries;
+        } catch (e) {
+            console.error(`Failed to fetch Aave APR in subgraph ${this.subgraphUrl}:`, e);
+            return {};
+        }
+    }
+}
+
+const wrappedAaveTokensV2Mainnet = {
```
timjrobinson: I think all of this config makes more sense in the network config files. Or something like:

```
aaveTokens: {
  [Network.Mainnet]: {
    tokensV2: {
      ...
    },
    wrappedTokensV2: {
      ...
    }, 
    tokensV3: {
      ...
    }
 }
}
```

(The underlying tokens should be defined somewhere else as they are not Aave specific)

Then having one function `getAaveAprHander(network: Network, version: number)` that pulls the correct config settings and returns a new `AaveAprHandler` with it, instead of defining a function for every network. 

In general network names should never need to be added to variable names, it's better to use a map. 

This will make it much easier to extend as we add more networks, as eventually we may have 50+ networks and this will become messy.

modules/pool/lib/apr-data-sources/ib-yield-apr-handlers/sources/default-apr-handler.ts
```diff
@@ -0,0 +1,183 @@
...
+
+export type DefaultAprHandlerConfig = {
+    tokens: string[];
+    url: string;
+    network: string;
+    scale?: number;
+    path?: string;
+};
+
+const vETHMainnet = '0x4bc3263eb5bb2ef7ad9ab6fb68be80e43b43801f';
```
timjrobinson: Like Aave, This could all be refactored into a map and all the handlers could be one function where you just pass in the network and it generates all handlers for that network on the fly and returns them in an array. Or you could pass in Network + Token and it returns a handler for that token if there is one defined.

modules/pool/lib/apr-data-sources/base-apr-handlers/base-apr-handlers.ts
```diff
@@ -0,0 +1,172 @@
...
+    getAprs(): Promise<{ [tokenAddress: string]: number }>;
+}
+
+export type TokenApr = {
+    val: number;
+    address: string;
+    group?: string;
+};
+
+// export const wrappedBoostedTokens = [
```
timjrobinson: comments to delete

timjrobinson (commented):
Nice work, this looks much cleaner!

modules/network/mainnet.ts
```diff
@@ -115,6 +116,201 @@ const mainnetNetworkData: NetworkData = {
...
+            sourceUrl: 'https://mainnet.gearbox.foundation/api/pools',
+            tokens: {
+                dDAI: '0x6cfaf95457d7688022fc53e7abe052ef8dfbbdba',
+                dUSDC: '0xc411db5f5eb3f7d552f9b8454b2d74097ccde6e3',
+            },
+        },
+        idle: {
+            sourceUrl: 'https://api.idle.finance/junior-rates/',
+            authorizationHeader:
+                'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJjbGllbnRJZCI6IkFwcDciLCJpYXQiOjE2NzAyMzc1Mjd9.L12KJEt8fW1Cvy3o7Nl4OJ2wtEjzlObaAYJ9aC_CY6M',
```
franzns: Does this need to be a secret?

modules/pool/lib/apr-data-sources/base-apr-handlers/sources/tessera-apr-handler.ts
```diff
@@ -0,0 +1,60 @@
...
+    tokens: {
+        [tokenName: string]: {
+            tesseraPoolAddress: string;
+            tokenAddress: string;
+        };
+    };
+    readonly group = 'TESSERA';
+
+    constructor(aprHandlerConfig: TesseraAprHandlerConfig) {
+        this.networkPrismaId = aprHandlerConfig.networkPrismaId;
```
franzns: Is this needed?

modules/pool/lib/apr-data-sources/ib-tokens-apr.service.ts
```diff
@@ -0,0 +1,85 @@
...
+        const tokenYieldPools = pools.filter((pool) => {
+            return pool.tokens.find((token) => {
+                return Array.from(aprs.keys())
+                    .map((key) => key.toLowerCase())
+                    .includes(token.address.toLowerCase());
+            });
+        });
+        for (const pool of tokenYieldPools) {
+            const totalLiquidity = pool.dynamicData?.totalLiquidity;
+            for (const token of pool.tokens) {
```
franzns, line 60: In theory, we should check whether the token has a rate provider set, as we only have an APR if there is a rate provider. We don't store that information in the backend though. Should probably add a TODO

modules/pool/lib/apr-data-sources/base-apr-handlers/base-apr-handlers.ts
```diff
@@ -0,0 +1,172 @@
...
+                handlers.push(handler);
+            }
+        }
+        return handlers;
+    }
+
+    async fetchAprsFromAllHandlers(): Promise<TokenApr[]> {
+        let aprs: { val: number; group: string; address: string }[] = [];
+        for (const handler of this.handlers) {
+            const fetchedResponse: { [key: string]: number } = await handler.getAprs();
```
franzns: Add try/catch here to make sure a faulty apr handler doesnt block all other services. Log to sentry with APR handlers name/ID

modules/pool/lib/apr-data-sources/base-apr-handlers/sources/tetu-apr-handler.ts
```diff
@@ -0,0 +1,40 @@
...
+    async getAprs() {
+        try {
+            const { data } = await axios.get(this.sourceUrl);
+            const json = data as { vault: string; apr: number }[];
+            const aprs = json
+                .filter(({ vault }) => Object.values(this.tokens).includes(vault.toLowerCase()))
+                .map((t) => [t.vault, t.apr / 100]);
+            return Object.fromEntries(aprs);
+        } catch (error) {
+            console.error('Failed to fetch Tetu APR:', error);
```
franzns: These should also log to sentry with some kind of ID of this handler. Seem we need to introduce a handler ID as well.

franzns: Above is for all APR handlders, not only this one.

modules/pool/lib/apr-data-sources/base-apr-handlers/sources/reaper-apr-handler.ts
```diff
@@ -0,0 +1,73 @@
+import { JsonRpcProvider } from '@ethersproject/providers';
+import { BigNumber, Contract } from 'ethers';
+import { abi } from './abis/reaperStrategy';
+
+import { AprHandler } from '../base-apr-handlers';
+
+class ReaperAprHandler implements AprHandler {
```
franzns: This does not implement the multistrategy APR fetching which needs a subgraph. I suggest to adapt the reaperHandlderConfig for this

modules/pool/lib/apr-data-sources/base-apr-handlers/sources/reaper-apr-handler.ts
```diff
@@ -0,0 +1,73 @@
...
+            });
+            const callsAprResults: BigNumber[] = Array(Object.keys(this.yieldTokens).length);
+            for (const [index, aprPromise] of contractCalls.entries()) {
+                callsAprResults[index] = await aprPromise;
+            }
+            const aprs = Object.keys(this.strategiesMap).map((tokenName, i) => {
+                return [this.yieldTokens[tokenName], callsAprResults[i].toNumber() / 1e4];
+            });
+
+            return Object.fromEntries(aprs);
```
franzns: I'd personally do this in one for-loop instead of multiple loops with `map` and `for`. I find this hard to read with all these `Object.` calls

modules/network/network-config-types.ts
```diff
@@ -100,6 +100,7 @@ export interface NetworkData {
         address: string;
         excludedFarmIds: string[];
     };
+    aprConfig: AprConfig;
```
franzns: rename this to baseIbAprConfig or similar. Needs to reflect this is only for the base APR config for any IB tokens

modules/pool/lib/apr-data-sources/apr-types.ts
```diff
@@ -1,3 +1,5 @@
+import { Dictionary } from 'lodash'
+
```
franzns: remove unused

modules/pool/lib/apr-data-sources/liquid-staked-base-apr.service.ts
```diff
@@ -1,6 +1,7 @@
 import axios from 'axios';
```
franzns, line 1: this whole class should be removed and integrated into the base-apr-handlers

modules/pool/lib/apr-data-sources/base-apr-handlers/base-apr-handlers.ts
```diff
@@ -0,0 +1,172 @@
...
+import { EulerAprHandler } from './sources/euler-apr-handler';
+import { GearboxAprHandler } from './sources/gearbox-apr-handler';
+import { IdleAprHandler } from './sources/idle-apr-handler';
+import { OvixAprHandler } from './sources/ovix-apr-handler';
+import { TesseraAprHandler } from './sources/tessera-apr-handler';
+import { TetuAprHandler } from './sources/tetu-apr-handler';
+import { TranchessAprHandler } from './sources/tranchess-apr-handler';
+import { AprConfig } from '../../../../network/network-config-types';
+
+export class BaseAprHandlers {
```
franzns: rename this to reflect this is only for IB and Linear APRs

modules/pool/lib/apr-data-sources/base-apr-handlers/sources/aave-apr-handler.ts
```diff
@@ -0,0 +1,98 @@
...
+        [assetName: string]: {
+            underlyingAssetAddress: string;
+            aTokenAddress: string;
+            wrappedTokens: {
+                [tokenName: string]: string;
+            };
+        };
+    };
+    subgraphUrl: string;
+    networkPrismaId: string;
```
franzns: For all handlers: I think these prismaId as well as networkId are not used anymore and can be removed.

modules/pool/lib/apr-data-sources/base-apr-handlers/sources/default-apr-handler.ts
```diff
@@ -0,0 +1,62 @@
...
+
+export class DefaultAprHandler implements AprHandler {
+    tokens: {
+        [tokenName: string]: string;
+    };
+    url: string;
+    path: string;
+    scale: number;
+    networkPrismaId: string;
+    group = 'DEFAULT';
```
franzns: this should be empty or undefined and only set if needed, otherwise if a pool as reth and wsteth apr it would be grouped under DEFAULT in the UI.

modules/pool/lib/apr-data-sources/base-apr-handlers/sources/default-apr-handler.ts
```diff
@@ -0,0 +1,62 @@
...
+
+    constructor(aprHandlerConfig: DefaultAprHandlerConfig) {
+        this.tokens = aprHandlerConfig.tokens;
+        this.url = aprHandlerConfig.sourceUrl;
+        this.networkPrismaId = aprHandlerConfig.networkPrismaId;
+        this.path = aprHandlerConfig.path ?? '';
+        this.scale = aprHandlerConfig.scale ?? 100;
+        this.group = aprHandlerConfig.group ?? 'DEFAULT';
```
franzns: this should be empty or undefined and only set if needed, otherwise if a pool as reth and wsteth apr it would be grouped under DEFAULT in the UI.

franzns (commented):
Thank you Luiz and Tim. This looks so much better! I have added comments where I found issues. Some comments I only added to one handler and wrote if they affect all of them.

In general, I like this approach way more! Go ahead and migrate all other IB APR services like this.

modules/network/mainnet.ts
lgahdl: To get this bearer token, all it needs is fill a typeform, I think it's not a sensitive datum

modules/pool/lib/apr-data-sources/base-apr-handlers/sources/reaper-apr-handler.ts
lgahdl: I think we could remove this Reaper Apr Handler and let the Reaper Service separated as it is, add the Reaper to the new service would move the Reaper rules (sftmx, multi and single strategy) that already exists into a new service and add complexity to the new service, what do you think?

modules/pool/lib/apr-data-sources/base-apr-handlers/sources/tessera-apr-handler.ts
lgahdl: Well noted! No, is not needed, will remove it

modules/pool/lib/apr-data-sources/base-apr-handlers/sources/reaper-apr-handler.ts
franzns: I would like to have the "old" reaper handler integrated into the new service you are creating. Adapt the reaperHandlerConfig to take singlestrategy vault addresses and multi-strategy vault address so you don't need to "hack" the way I did it.

modules/pool/lib/apr-data-sources/ib-tokens-apr.service.ts
```diff
@@ -0,0 +1,140 @@
...
+                                },
+                            }),
+                        );
+                    }
+                }
+            } else {
+                // If pool is Linear
+                if (!isLinear) {
+                    continue;
+                }
```
franzns: not needed

modules/pool/lib/apr-data-sources/ib-tokens-apr.service.ts
```diff
@@ -0,0 +1,140 @@
...
+        for (const pool of tokenYieldPools) {
+            if (!pool.dynamicData) {
+                continue;
+            }
+            const totalLiquidity = pool.dynamicData?.totalLiquidity;
+            if (!totalLiquidity) {
+                continue;
+            }
+            const isLinear = pool.type === 'LINEAR' && pool.linearData !== null;
+            if (!isLinear) {
```
franzns: this is essentially to find out whether its `IB_YIELD` or `LINEAR_BOOSTED`. That should be able to do it easier without duplicating all this code

modules/pool/lib/apr-data-sources/ib-linear-apr-handlers/sources/reaper-crypt-apr-handler.ts
```diff
@@ -0,0 +1,141 @@
...
+            if (!token) {
+                return acc;
+            }
+            let tokenApr = parseFloat(apr) / APR_PERCENT_DIVISOR;
+            if (token.isSftmX) {
+                tokenApr = tokenApr * (1 + sFTMxBaseApr);
+            }
+            if (token.isWstETH) {
+                tokenApr = tokenApr * (1 + this.wstETHBaseApr);
+            }
```
franzns: is there a possibility to make this more general? Like if any of the IB-token-apr handlers for linear pools have an IB token, this needs to apply.

modules/pool/lib/apr-data-sources/reaper-crypt-apr.service.ts
```diff
@@ -1,181 +0,0 @@
-import { isSameAddress } from '@balancer-labs/sdk';
```
franzns, line 1: remove alltogether

modules/pool/lib/apr-data-sources/fantom/yearn-vault-apr.service.ts
```diff
@@ -1,59 +0,0 @@
-import { PoolAprService } from '../../../pool-types';
```
franzns, line 1: remove file

modules/pool/lib/apr-data-sources/apr-types.ts
franzns: still there :)

franzns (commented):
Thanks, still a few things. Also, I think Optimism is missing reaper single strat configs (maybe also fantom).

Also, there are two reaper apr handlers, I believe only one is needed.

timjrobinson:
I've been testing this out with https://github.com/balancer/frontend-v2/pull/4068 running it all locally, and found a few issues:

- All pools are missing BAL rewards

### Base

- USDbC rewards on cbETH/WETH (`0xfb4c2e6e6e27b5b4a07a36360c89ede29bb3c9b6000000000000000000000026`) and USDbC/axlUSDC (`0xe58ca65f418d4121d6c70d4c133e60cf6fda363c000000000000000000000013`) pools on base are way too low (like orders of magnitude lower than expected, seems like a scaling issue maybe because the reward token has 6 decimals rather than 18)
- Pool liquidity values are incorrect (different to prod)
- GOLD rewards APR are much lower than prod (40% vs 60%)
- There's no APR on `0x18f150c43598cd822d39711c55bd90407a8b8ad700020000000000000000003b` when in prod it has 12% APR - rewards from the subpool `0x2db50a0e0310723ef0c2a165cb9a9f80d772ba2f00020000000000000000000d`

### Ethereum

- `0x1e19cf2d73a72ef1332c882f20534b6519be0276000200000000000000000112` is missing rETH APR while `0x42ed016f826165c2e5976fe5bc3df540c5ad0af700000000000000000000058b` has the token APR's including rETH

For some things (like missing BAL rewards) I'm thinking maybe I haven't run the right jobs to get the data needed?

It would be good to get this merged to canary or hosted somewhere where it can continually update with the latest data and  anyone compare current Balancer prod with pulling pools from the beets API and report any differences.

franzns:
Thanks for testing further! We are in the final phase to test everything locally. Next phase is to merge to canary and have it running there to investigate differences between prod and canary, like you already did.

The missing/wrong BAL rewards is something we already noticed and @gmbronco is looking into that separately.

lgahdl:
> * 0x1e19cf2d73a72ef1332c882f20534b6519be0276000200000000000000000112
This rETH token included in the APR's is the stafi rETH (0x9559aaa82d9649c7a7b220e7c461d2e74c9a3593), not the rETH you're referring to (0xae78736cd615f374d3085123a210448e74fc6393), that's why this pool is not getting the rETH APR, thanks for seeing that.
 I'm not sure if the source we're using for the stafi rETH could be used for the rETH, but I can add it

lgahdl:
> I've been testing this out with [balancer/frontend-v2#4068](https://github.com/balancer/frontend-v2/pull/4068) running it all locally, and found a few issues:
> 
> * All pools are missing BAL rewards
> 
> ### Base
> * USDbC rewards on cbETH/WETH (`0xfb4c2e6e6e27b5b4a07a36360c89ede29bb3c9b6000000000000000000000026`) and USDbC/axlUSDC (`0xe58ca65f418d4121d6c70d4c133e60cf6fda363c000000000000000000000013`) pools on base are way too low (like orders of magnitude lower than expected, seems like a scaling issue maybe because the reward token has 6 decimals rather than 18)
> * Pool liquidity values are incorrect (different to prod)
> * GOLD rewards APR are much lower than prod (40% vs 60%)
> * There's no APR on `0x18f150c43598cd822d39711c55bd90407a8b8ad700020000000000000000003b` when in prod it has 12% APR - rewards from the subpool `0x2db50a0e0310723ef0c2a165cb9a9f80d772ba2f00020000000000000000000d`
> 
> ### Ethereum
> * `0x1e19cf2d73a72ef1332c882f20534b6519be0276000200000000000000000112` is missing rETH APR while `0x42ed016f826165c2e5976fe5bc3df540c5ad0af700000000000000000000058b` has the token APR's including rETH
> 
> For some things (like missing BAL rewards) I'm thinking maybe I haven't run the right jobs to get the data needed?
> 
> It would be good to get this merged to canary or hosted somewhere where it can continually update with the latest data and anyone compare current Balancer prod with pulling pools from the beets API and report any differences.

This rETH token included in the APR's is the stafi rETH (0x9559aaa82d9649c7a7b220e7c461d2e74c9a3593), not the rETH you're referring to (0xae78736cd615f374d3085123a210448e74fc6393), that's why this pool(0x1e19cf2d73a72ef1332c882f20534b6519be0276000200000000000000000112) is not getting the rETH APR, thanks for seeing that.
 I'm not sure if the source we're using for the stafi rETH could be used for the rETH, but I can add it
 Note.: The rETH APR of the pool 0x42ed016f826165c2e5976fe5bc3df540c5ad0af700000000000000000000058b is probably not from this service

PR #398, opened 2023-07-06
Chore/install viem

agualis:
- Installs viem 
- Ensures that the app is running in node 18  (viem requirement)
- Updates tsconfig "target": "ES2020" (viem requirement)


agualis:
Decided not to start using viem for now.

PR #432, opened 2023-08-09
Fix/gauge controller helper

agualis:


modules/vebal/vebal-voting-list.service.ts
```diff
@@ -114,14 +114,14 @@ export class VeBalVotingListService {
...
-    async syncVotingGauges() {
+    public async syncVotingGauges() {
         const onchainGaugeAddresses = await this.votingGauges.getVotingGaugeAddresses();
 
         return this.sync(onchainGaugeAddresses);
     }
 
     async sync(votingGaugeAddresses: string[]) {
-        const chunks = chunk(votingGaugeAddresses, 100);
+        const chunks = chunk(votingGaugeAddresses, 50);
```
agualis, line 128: @franzns I found that the new `gaugeControllerHelperAbi` was failing when doing a multicall with 100 elements. This fixes the problem but we could do something better. Could you help me with this?

franzns: think it's totally fine like this

modules/vebal/vebal-voting-list.service.ts
```diff
@@ -2,14 +2,18 @@ import { prisma } from '../../prisma/prisma-client';
...
+import { oldVeBalAddress, specialVotingGaugeAddresses } from './special-pools/special-voting-gauge-addresses';
 import { getVeVotingGauge, veGauges, vePools } from './special-pools/ve-pools';
 import { hardCodedPools } from './special-pools/hardcoded-pools';
 import { GqlVotingPool } from '../../schema';
 
 export class VeBalVotingListService {
     constructor(private votingGauges = new VotingGaugesRepository()) {}
 
+    /*
+        👀 This methods id used by veBalGetVotingList resolver that is consumed by some partners 👀
```
franzns: can you remove these eyes please?

modules/vebal/vebal.resolvers.ts
```diff
@@ -12,6 +12,16 @@ const resolvers: Resolvers = {
         veBalGetTotalSupply: async (parent, {}, context) => {
             return veBalService.getVeBalTotalSupply();
         },
+        /*
+            👀 This endpoint is consumed by some partners 👀
```
franzns: same

PR #451, opened 2023-09-08
IB Yield APR Service - Adding new tokens and handlers

lgahdl:
Adding new tokens
Creating 2 new apr Handlers: Bloom and Maker;

### DIFF
--- a/modules/network/avalanche.ts
+++ b/modules/network/avalanche.ts
@@ -174,7 +174,7 @@ const avalancheNetworkData: NetworkData = {
                 tokenAddress: '0xa25eaf2906fa1a3a13edac9b9657108af7b703e3',
                 sourceUrl: 'https://ceres.gogopool.com',
                 path: 'ggAVAXMonthlyInterestMonth.value',
-                scale: -0.8333,
+                scale: -8.3333,
             },
         },
     },
--- a/modules/network/avalanche.ts
+++ b/modules/network/avalanche.ts
@@ -174,6 +174,9 @@ const avalancheNetworkData: NetworkData = {
                 tokenAddress: '0xa25eaf2906fa1a3a13edac9b9657108af7b703e3',
                 sourceUrl: 'https://ceres.gogopool.com',
                 path: 'ggAVAXMonthlyInterestMonth.value',
+                // According to solarcurve, the AVAX Monthly Interest must be multiplied by -12 to represent the APR in normal scale, for example, if the monthly interest is -0,15, the APR would be -0,15 * -12 = 1,8%.
+                // How this -12 became -8,333? It's because the scale parameter is used to divide the number, and the final apr percentage is in decimal format (1,8% = 0,018), so if:
+                // M * -12 = A (M is monthly rate and A is APR) => (M/x) = (A/100) => (A / -12x) = (A / 100) [replacing M by A/-12] => x = 100/-12 = -8,33333
                 scale: -8.3333,
             },
         },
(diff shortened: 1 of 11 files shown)


modules/network/avalanche.ts
```diff
@@ -158,6 +158,25 @@ const avalancheNetworkData: NetworkData = {
...
+            yyAVAX: {
+                tokenAddress: '0xf7d9281e8e363584973f946201b82ba72c965d27',
+                sourceUrl: 'https://staging-api.yieldyak.com/yyavax',
+                path: 'yyAVAX.apr',
+            },
+            ggAVAX: {
+                tokenAddress: '0xa25eaf2906fa1a3a13edac9b9657108af7b703e3',
+                sourceUrl: 'https://ceres.gogopool.com',
+                path: 'ggAVAXMonthlyInterestMonth.value',
+                scale: -0.8333,
```
franzns: is there an explanation for this number?

modules/network/polygon.ts
```diff
@@ -204,6 +204,18 @@ const polygonNetworkData: NetworkData = {
                 path: 'value',
                 isIbYield: true,
             },
+            overnightUSDPlus: {
+                tokenAddress: '0x5d9d8509c522a47d9285b9e4e9ec686e6a580850',
+                sourceUrl: 'https://api.overnight.fi/optimism/usd+/fin-data/avg-apr/week',
+                path: 'value',
+                group: 'OVERNIGHT',
+            },
+            overnightstUSDPlus: {
```
franzns, line 213: same token different address?

modules/pool/lib/apr-data-sources/ib-linear-apr-handlers/sources/bloom-apr-handler.ts
```diff
@@ -0,0 +1,34 @@
...
+        const aprs: { [p: string]: { apr: number; isIbYield: boolean } } = {};
+        for (const { address, feedAddress, isIbYield } of Object.values(this.tokens)) {
+            try {
+                const feedContract = getContractAt(feedAddress, bloomBpsFeed);
+                const currentRate = await feedContract.currentRate();
+                if (!currentRate) {
+                    continue;
+                }
+                const tokenApr = (Number(currentRate) - 10000) / 10000;
+                aprs[address] = { apr: tokenApr, isIbYield: isIbYield ?? false };
```
franzns, line 26: why do `isIbYield: isIbYield ?? false` ? just set it to false

modules/network/polygon.ts
lgahdl: No, this is the stUSD+, the other one is the USD+

modules/pool/lib/apr-data-sources/ib-linear-apr-handlers/sources/bloom-apr-handler.ts
lgahdl: I made like this because I didn't know if this apr handler would be used for non-ib-yield tokens in the future

modules/network/avalanche.ts
lgahdl: Yeah, according with solarcurve, the AVAX Monthly Interest must be multiplied by -12 to represent the APR in normal scale, for example, if the monthly interest is -0,15, the APR would be -0,15 * -12 = 1,8%, my understanding is that this number represents what the owner pays as interest, if he receives, it's negative, but I'll confirm this last statement with him.
Now, how this -12 became -8,333 (sorry, the -0,8333 was wrong, already fixed), it's because the scale parameter is used to divide the number, and the final apr percentage is in decimal format (1,8% = 0,018), so if 
M * -12 = A (M is monthly rate and A is APR) => (M/x) = (A/100) => (A / -12x) = (A / 100) [replacing M by A/-12] => x = 100/-12 = -8,33333

franzns: can you add this as a comment as well please?

modules/network/polygon.ts
franzns: ah didn't catch that 👍

modules/network/avalanche.ts
lgahdl: Of course

PR #453, opened 2023-09-12
aprs: adding working supply

gmbronco:
Adding working supply to the BAL APR calculations. There are still differences on mainnet after checking with the current SDK APRs.


modules/pool/lib/staking/gauge-staking.service.ts
```diff
@@ -180,40 +190,84 @@ export class GaugeStakingService implements PoolStakingService {
...
-                };
-            } else {
-                response[childChainGauge] = {
-                    version: 1,
-                    rate: '0.0',
-                };
+        const results = (await multicall.execute()) as {
+            [address: string]: {
+                inflationRate: BigNumber | undefined,
+                workingSupply: BigNumber
```
franzns: v1 gauges don't have a workingSupply

modules/pool/lib/staking/gauge-staking.service.ts
```diff
@@ -180,40 +190,84 @@ export class GaugeStakingService implements PoolStakingService {
...
+
+    /**
+     * Get the inflation rate for all the gauges on mainnet.
+     * Gauges on mainnet use the inflation rate from token admin contract
+     * and relative weight from the gauge contract.
+     *
+     * @param gaugeAddresses 
+     * @returns 
+     */
+    private async getMainnetGaugeRates(gaugeAddresses: string[]): Promise<{ [gaugeAddress: string]: GaugeRate }> {
```
franzns: are there also v1 and v2 mainnet gauges that need separate handling?

modules/pool/lib/staking/gauge-staking.service.ts
```diff
@@ -180,40 +190,84 @@ export class GaugeStakingService implements PoolStakingService {
...
+        const now = Math.round(new Date().getTime() / 1000);
+        const version = 2; // On Mainnet BAL is always distributed directly to gauges
+        const { gaugeControllerAddress } = networkContext.data;
+        const abi = [
+            'function working_supply() view returns (uint256)',
+            'function gauge_relative_weight(address) view returns (uint256)'
+        ];
+        const multicall = new Multicaller3(abi);
```
franzns: the gaugeController ABI is present under modules/vebal/abi . I'd suggest to move to modules/web3/abi and import from there. If we do this inline like this, I'm sure at some point the gaugeController changes and these functions are different

modules/pool/lib/staking/gauge-staking.service.ts
```diff
@@ -180,40 +190,84 @@ export class GaugeStakingService implements PoolStakingService {
...
+            multicall.call(`${address}.weight`, gaugeControllerAddress!, 'gauge_relative_weight', [address], true);
+            multicall.call(`${address}.workingSupply`, address, 'working_supply', [], true);
+        });
+        
+        const gaugeData = await multicall.execute() as { [address: string]: { weight: BigNumber, workingSupply: BigNumber } };
+        const inflationRate = emissions.weekly(now) / 604800; // BAL inflation rate per second
+
+        const weightedRates = _.mapValues(gaugeData, ({ weight }) => {
+            if (weight.eq(0)) return 0;
+            return inflationRate * Number(formatUnits(weight));
```
franzns: I usually to parseFloat(), is there a difference?

modules/pool/lib/staking/gauge-staking.service.ts
```diff
@@ -180,40 +190,84 @@ export class GaugeStakingService implements PoolStakingService {
...
+        ];
+        const multicall = new Multicaller3(abi);
+
+        gaugeAddresses.forEach((address) => {
+            multicall.call(`${address}.weight`, gaugeControllerAddress!, 'gauge_relative_weight', [address], true);
+            multicall.call(`${address}.workingSupply`, address, 'working_supply', [], true);
+        });
+        
+        const gaugeData = await multicall.execute() as { [address: string]: { weight: BigNumber, workingSupply: BigNumber } };
+        const inflationRate = emissions.weekly(now) / 604800; // BAL inflation rate per second
```
franzns: Couldn't we just use
```
/**
     * @notice Returns the current inflation rate of BAL per second
     */
    function getInflationRate() external view returns (uint256) {
        return _rate;
    }

```
from the tokenAdmin contract?

franzns (changes requested):
Overall good, thanks. A few comments in the code, main two issues:
1) working supply doesnt exist on v1 gauges
2) bal inflation rate could be queried from the tokenAdmin contract?

modules/pool/lib/staking/gauge-staking.service.ts
gmbronco: no difference, changed to parseFloat

modules/pool/lib/staking/gauge-staking.service.ts
gmbronco: yes we could. should I add it this call? it's returning exactly the same data.

modules/pool/lib/staking/gauge-staking.service.ts
gmbronco: changed to the ABI, we might need to do some work anyways when it changes.

modules/pool/lib/staking/gauge-staking.service.ts
gmbronco: no, it's too much branching. i refactored it, but it's still quite hard to read.

modules/pool/lib/staking/gauge-staking.service.ts
gmbronco: true, it will be used only in V2 and mainnet.

modules/pool/lib/staking/gauge-staking.service.ts
franzns: that would be great if you could add it. I like to have an onchain query rather than local calculation, saves a lot of code and less room for error

modules/pool/lib/staking/gauge-staking.service.ts
```diff
@@ -1,221 +1,252 @@
...
-                    rate: '0.0',
-                };
+        // Get onchain data for reward tokens
+        for (const gauge of gauges) {
+            for (const token of gauge.tokens ?? []) {
+                const [address] = token.id.toLowerCase().split('-');
+                if (gauge.version === 1) {
+                    this.rewardsMulticallerV1.call(`${gauge.id}.rewardData.${address}`, gauge.id, 'reward_data', [address], true);
+                } else {
+                    this.rewardsMulticallerV2.call(`${gauge.id}.rewardData.${address}`, gauge.id, 'reward_data', [address], true);
```
franzns: why the differentiation? It looks exactly the same.

modules/pool/lib/staking/gauge-staking.service.ts
```diff
@@ -1,221 +1,252 @@
...
+                    : (parseFloat(formatUnits(balData[gaugeAddress].weight!)) * totalBalRate).toFixed(18), // mainnet case
+                workingSupply: balData[gaugeAddress]?.workingSupply ? formatUnits(balData[gaugeAddress].workingSupply!) : '0',
+            })),
+            ...Object.keys(rewardsData).map((gaugeAddress) => [ // L2 V1 case, includes tokens other than BAL
+                ...Object.keys(rewardsData[gaugeAddress].rewardData).map((address) => ({
+                    id: `${address}-${gaugeAddress}`.toLowerCase(),
+                    rewardPerSecond: rewardsData[gaugeAddress].rewardData[address]?.period_finish
+                        && rewardsData[gaugeAddress].rewardData[address]?.period_finish!.toNumber() > now
+                        && formatUnits(rewardsData[gaugeAddress].rewardData[address].rate!)
+                        || '0.0',
```
franzns: Why would this be a string and not a boolean? What do I miss?

modules/pool/lib/staking/gauge-staking.service.ts
```diff
@@ -1,221 +1,252 @@
...
+        // Format onchain rates for all the rewards
+        const onchainRates = [
+            ...Object.keys(balData).map((gaugeAddress) => ({
+                id: `${this.balAddress}-${gaugeAddress}`.toLowerCase(),
+                rewardPerSecond: balData[gaugeAddress]?.rate
+                    ? formatUnits(balData[gaugeAddress].rate!) // L2 V2 case
+                    : (parseFloat(formatUnits(balData[gaugeAddress].weight!)) * totalBalRate).toFixed(18), // mainnet case
+                workingSupply: balData[gaugeAddress]?.workingSupply ? formatUnits(balData[gaugeAddress].workingSupply!) : '0',
+            })),
+            ...Object.keys(rewardsData).map((gaugeAddress) => [ // L2 V1 case, includes tokens other than BAL
```
franzns, line 237: doesnt v1 gauges include tokens including BAL?

prisma/schema.prisma
```diff
@@ -470,6 +470,7 @@ model PrismaPoolStakingGauge {
     rewards             PrismaPoolStakingGaugeReward[]
     status              PrismaPoolStakingGaugeStatus    @default(ACTIVE)
     version             Int                             @default(1)
+    workingSupply       String                          @default("0.0")
```
franzns, line 474: this needs to be added to pool.prisma in the pool module

modules/pool/lib/staking/gauge-staking.service.ts
```diff
@@ -1,221 +1,252 @@
+/**
+ * Supports calculation of BAL rewards sent to gauges. Balancer setup has 3 types of gauges:
```
franzns: also supports any thirdparty reward tokens

modules/pool/lib/staking/gauge-staking.service.ts
gmbronco: not sure I understand, do you mean the logic? It translates to: if period_finish was fetched and it's in the future then use the formatted rate otherwise set it to 0.

gmbronco: zeros are filtered out later on, so we don't have store empty values in the DB

modules/pool/lib/staking/gauge-staking.service.ts
gmbronco: this import seems to be breaking in the github workflow, would you know why?

modules/pool/lib/staking/gauge-staking.service.ts
gmbronco: The return values from the reward_data function vary between the old and new L2 gauges.

modules/pool/lib/apr-data-sources/ve-bal-gauge-apr.service.ts
```diff
@@ -92,12 +92,22 @@ export class GaugeAprService implements PoolAprService {
                     const aprItemId = `${pool.id}-${rewardTokenDefinition.symbol}-apr`;
                     const aprRangeId = `${pool.id}-bal-apr-range`;
 
+                    // Only 40% of LP token staked accrue emissions, totalSupply = workingSupply * 2.5
+                    const workingSupply = (parseFloat(preferredStaking.gauge.workingSupply) + 0.4) / 0.4;
```
franzns: you already get the working supply, why `+0.4)/0.4` ?

franzns (approved):

PR #456, opened 2023-09-15
Adding type and version to graphQL pool models:

lgahdl:
- GqlPoolBase;
- GqlPoolWeighted;
- GqlPoolStable;
- GqlPoolPhantomStable;
- GqlPoolMetaStable;
- ..and all the other Pools of GqlPoolUnion type

### DIFF
--- a/modules/pool/pool.gql
+++ b/modules/pool/pool.gql
@@ -97,6 +97,7 @@ interface GqlPoolBase {
     #fields that never change after creation
     id: ID!
     chain: GqlChain!
+    type: String!
     name: String!
     symbol: String!
     address: Bytes!
@@ -215,6 +216,7 @@ type GqlPoolWeighted implements GqlPoolBase {
     tokens: [GqlPoolTokenUnion!]!
     nestingType: GqlPoolNestingType!
     staking: GqlPoolStaking
+    type: String!
 }
 
 type GqlPoolGyro implements GqlPoolBase {
@@ -240,6 +242,7 @@ type GqlPoolGyro implements GqlPoolBase {
     tokens: [GqlPoolTokenUnion!]!
     nestingType: GqlPoolNestingType!
     staking: GqlPoolStaking
+    type: String!
 }
 
 type GqlPoolLiquidityBootstrapping implements GqlPoolBase {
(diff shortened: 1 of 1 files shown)


franzns:
What is the reason for this?

lgahdl:
> What is the reason for this?

The type of the pool is a required parameter for the b-sdk:

```ts
export type PoolState = {
    id: Address;
    address: Address;
    type: string;
    tokens: {
			address: Address;
			decimals: number;
			index: number;
}[];
```

franzns:
you can query the type via `__typename` in the pools Object which returns you the name of the GQL model (e.g. GqlPoolWeighted). Is that sufficient?

lgahdl:
> you can query the type via `__typename` in the pools Object which returns you the name of the GQL model (e.g. GqlPoolWeighted). Is that sufficient?

It would work, but, for example, to check if a Gyro Pool is Gyro2, Gyro3 or GyroE, it would need to look what factory created the pool, because all of them has the __typename "GqlPoolGyro", in my opinion it's easier to use this "type" parameter, and I'm thinking about the same thing for the "version", that is also not in the models, maybe it would be good to add "version" as well in the gql models, because some pool versions has different token ordering than others, what do you think?

franzns:
Ok, makes sense. But it doesnt work like this. 
1) You only need to add it to `GqlPoolBase`, as all other extend that model
2) You also need to populate that field in `pool-gql-loader.service.ts`
3) Might also make sense to add it to `GqlPoolMinimal`

You want to give that a shot?

lgahdl:
Hi @franzns! I made the changes to add "version" to all types, and I already tested for poolGetPools and poolGetPool methods, it's working fine. It wasn't necessary to populate these fields on "pool-gql-loader.service.ts", they come automatically populated because they're already part of "GqlPoolMinimal" and "GqlPoolBase" at schema.ts, so these fields don't need to be included manually, just like "id", "name", "symbol", and other base fields.

About the "1. You only need to add it to GqlPoolBase, as all others extend that model", this type of inheritance is not available in the basic graphQL, the interface "implements" keyword requires the child type to have inherited fields written(like id, name, chain, and others), but there are some packages that make possible to create directives and implement this type of functionality for graphQL, I can search further if you like, the most trustable one seems to be "@graphql-tools/utils".

StackOverflow Question about graphQL inheritance: https://stackoverflow.com/questions/47523384/how-to-inherit-or-extend-typedefs-in-graphql
Implementing this Inheritance in graphQL: https://nelsondominguez.hashnode.dev/inheritance-in-graphql-when-and-how-to-use-it
Package that implements graphQL directives: https://the-guild.dev/graphql/tools/docs/schema-directives

franzns:
Thanks luiz!

PR #457, opened 2023-09-18
Add FX pool type

timjrobinson:
Adding two pool types needed by the Balancer frontend for filtering by pool type.


franzns:
Will create FX pool type support in a different branch

PR #482, opened 2023-10-16
Add chains array param to user balance query

danielmkm:
Add support for fetching user balances from multiple chains at once. As the balances leverage the token prices, we also implement support for multi chain token price fetching.

### DIFF
--- a/modules/network/network-config.ts
+++ b/modules/network/network-config.ts
@@ -8,6 +8,8 @@ import { gnosisNetworkConfig } from './gnosis';
 import { zkevmNetworkConfig } from './zkevm';
 import { avalancheNetworkConfig } from './avalanche';
 import { baseNetworkConfig } from './base';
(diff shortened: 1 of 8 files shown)


danielmkm:
~TODO: need to add support for fetching prices for multiple chains~

gmbronco (approved):
lgtm – i just added a small change making it default to all chains if no chains are provided and an address filter, might be slightly easier to query.

danielmkm:
> lgtm – i just added a small change making it default to all chains if no chains are provided and an address filter, might be slightly easier to query.

Thanks for having a look @gmbronco. The way you did `allChains` will include fantom and optimism for balancer. You'd want to scope it to the project.

gmbronco:
we changed the default fallback to the chainId passed in the header, so it's backwards compatibile.

PR #483, opened 2023-10-17
Add chains to tokens and token prices queries

danielmkm:


gmbronco:
added all chains as a default, would that be helpful?

PR #485, opened 2023-10-17
Config refactor avoiding circular dep issues

gmbronco:


modules/network/arbitrum/index.ts
```diff
@@ -0,0 +1,43 @@
+import { JsonRpcProvider } from '@ethersproject/providers';
+import { data } from './data';
+import { workerJobs } from './workers';
+import { createServices } from './services';
```
danielmkm: can we prefix these with the network name. I know its an opinion thing here, but exposing 10 versions of `data`, `workerJobs` and `createServices` to the global namespace is messy.

modules/network/arbitrum/index.ts
```diff
@@ -0,0 +1,43 @@
...
+
+let services: ReturnType<typeof createServices>;
+
+export class arbitrumNetworkConfig {
```
danielmkm: Can you capitalize the first letter now that these are classes

danielmkm (changes requested):
Couple of small syntax changes if you don't mind, otherwise looks good! Thanks for the work here.

modules/network/arbitrum/services.ts
```diff
@@ -0,0 +1,35 @@
...
+import { SwapsPriceHandlerService } from '../../token/lib/token-price-handlers/swaps-price-handler.service';
+import { UserSyncGaugeBalanceService } from '../../user/lib/user-sync-gauge-balance.service';
+import { GithubContentService } from '../../content/github-content.service';
+import { gaugeSubgraphService } from '../../subgraphs/gauge-subgraph/gauge-subgraph.service';
+import { CoingeckoPriceHandlerService } from '../../token/lib/token-price-handlers/coingecko-price-handler.service';
+import { coingeckoService } from '../../coingecko/coingecko.service';
+import { IbTokensAprService } from '../../pool/lib/apr-data-sources/ib-tokens-apr.service';
+import { arbitrumNetworkData as data } from './data';
+
+export const arbitrumCreateServices = () => ({
```
franzns, line 18: they would need to be typed to NetworkService I guess?

modules/network/arbitrum/index.ts
```diff
@@ -0,0 +1,10 @@
+import { JsonRpcProvider } from '@ethersproject/providers';
+import { arbitrumNetworkData as data } from './data';
+import { arbitrumWorkerJobs as workerJobs } from './workers';
+export * from './services';
+
+export class ArbitrumNetworkConfig {
```
franzns, line 5: also missing the type?

modules/network/arbitrum/workers.ts
```diff
@@ -0,0 +1,99 @@
...
+import { DeploymentEnv } from '../network-config-types';
+
+/*
+For sub-minute jobs we set the alarmEvaluationPeriod and alarmDatapointsToAlarm to 1 instead of the default 3. 
+This is needed because the minimum alarm period is 1 minute and we want the alarm to trigger already after 1 minute instead of 3.
+
+For every 1 days jobs we set the alarmEvaluationPeriod and alarmDatapointsToAlarm to 1 instead of the default 3. 
+This is needed because the maximum alarm evaluation period is 1 day (period * evaluationPeriod).
+*/
+export const arbitrumWorkerJobs = [
```
franzns: also type

modules/network/mainnet/index.ts
```diff
@@ -0,0 +1,9 @@
+import { JsonRpcProvider } from '@ethersproject/providers';
+import { mainnetNetworkData as data } from './data';
+import { mainnetWorkerJobs as workerJobs } from './workers';
+
```
franzns, line 4: missing export

gmbronco: actually it's not needed. i'm removing from other networks as well.

modules/network/arbitrum/workers.ts
gmbronco: adding the type

modules/network/arbitrum/services.ts
gmbronco: i'm thinking how helpful that would be, since it's a static config adding a type only helps when someone forgets a required service, but even then it would get picked up by inference.

modules/network/arbitrum/index.ts
gmbronco: that's done

modules/network/arbitrum/index.ts
gmbronco: done

gmbronco:
hey @danielmkm - i think all the changes are there, let me know if you want to get anything else updated.

gmbronco:
I'll close this PR for now, since it's not solving the underlying issue fully. The issue is related to the networkContext and we will be refactoring that for specific cases.

PR #496, opened 2023-10-26
add featured pools

mendesfabio:
hey, folks! this is my first PR to the API 🎉 I talked to @danielmkm and he suggested me to work on this, some context: 

> Gareth made this [repository](https://github.com/balancer/metadata/blob/main/pools/featured.json) to house metadata for pools. Would be great is if the API fetched this JSON and then merged it with the actual pool data, and returned it in the same format as is returned by the `poolGetPools` query `GqlPoolMinimal`. Lastly, expose a query named `poolGetFeaturedPools` and return an array of featured pools.

He suggested the following model for a featured pool:

```
{
    "id": ID!,
    "primary": Boolean!,
    "imageUrl": String!,
    "pool": GqlPoolMinimal!
}
```

Happy to hear any feedback!


gmbronco:
Hey! Thanks for the pull request; nice one! I would suggest caching the GitHub data in the database so that we don't depend on fetching from an external URL.

modules/pool/lib/pool-gql-loader.service.ts
```diff
@@ -66,6 +68,20 @@ export class PoolGqlLoaderService {
         return pools.map((pool) => this.mapToMinimalGqlPool(pool));
     }
 
+    public async getFeaturedPools(): Promise<GqlFeaturedPool[]> {
+        const poolsMetadata = await poolMetadataService.getPoolsMetadata();
+        const poolIds = poolsMetadata.map((pool) => pool.id);
+
+        const pools = await this.getPools({ where: { idIn: poolIds } });
```
franzns: This fetches the pools from the subgraph but we want to avoid that. Pools are synced to the database. You can query them like this:

```
    const pools = await prisma.prismaPool.findMany({
        where: { id: { in: poolIds }, chain: networkContext.chain },
    });
```

This returns the prismaPool object, not sure if you need additional data. Also, you need to pass it the chain you request the pools for via network context. Or better, your gql resolver should take a chain array as argument and you query the pools for those chains.

franzns (changes requested):
In general, the code works like this but it should be integrated in the content service that we already have. Check how we expose featured pools here:
https://github.com/beethovenxfi/beethovenx-backend/blob/v3-main/modules/pool/lib/pool-gql-loader.service.ts#L93

You should be able to extend the github content service for Balancer:  https://github.com/beethovenxfi/beethovenx-backend/blob/v3-main/modules/content/github-content.service.ts#L161

You can take a look at how we did it for the featured pools for beets:  https://github.com/beethovenxfi/beethovenx-backend/blob/v3-main/modules/content/github-content.service.ts#L161

You can also try and persist the data into the database as jarek suggested, but can also leave that for another day.

Let me know if you have any questions.

modules/pool/lib/pool-gql-loader.service.ts
mendesfabio: https://github.com/beethovenxfi/beethovenx-backend/pull/496/commits/9b5f0f6b5f793991178215c93789c651c7996d7c

modules/pool/pool.service.ts
```diff
@@ -114,7 +114,7 @@ export class PoolService {
         return this.poolSwapService.getJoinExits(args);
     }
 
-    public async getFeaturedPoolGroups(): Promise<GqlPoolFeaturedPoolGroup[]> {
+    public async getFeaturedPoolGroups(chains: Chain[]): Promise<GqlPoolFeaturedPoolGroup[]> {
         const cached: GqlPoolFeaturedPoolGroup[] = await this.cache.get(
```
mendesfabio: @franzns I'm in doubt about how to re-write this and `poolGqlLoaderService.getFeaturedPoolGroups()` functions given `chains`. Should we try to get it from cache for every network and for the ones it doesn't find we get it from `poolGqlLoaderService`?

franzns: It's fine to simply get rid of this cache since we are caching on Stellate. This was from a time before that.

modules/content/github-content.service.ts
```diff
@@ -159,7 +166,15 @@ export class GithubContentService implements ContentService {
     }
     async syncPoolContentData(): Promise<void> {}
     async getFeaturedPoolGroups(): Promise<HomeScreenFeaturedPoolGroup[]> {
```
franzns: this would need to be changed to take the chain (or chains) as an argument. Also for the sanity content service.

modules/pool/pool.resolvers.ts
```diff
@@ -49,8 +49,8 @@ const balancerResolvers: Resolvers = {
             }
             return poolService.getPoolJoinExits(args);
         },
-        poolGetFeaturedPoolGroups: async (parent, args, context) => {
-            return poolService.getFeaturedPoolGroups();
+        poolGetFeaturedPoolGroups: async (parent, { chains }, context) => {
+            return poolService.getFeaturedPoolGroups(chains);
```
franzns, line 59: here we would want to do the same that we do e.g. `poolGetAllPoolsSnapshots` below. We have a migration period where we still support both, the header chainId as well as the chains param

modules/content/github-content.service.ts
```diff
@@ -5,8 +5,15 @@ import { prisma } from '../../prisma/prisma-client';
...
 import { ContentService, HomeScreenFeaturedPoolGroup, HomeScreenNewsItem } from './content-types';
 
+const POOLS_METADATA_URL = "https://raw.githubusercontent.com/balancer/metadata/main/pools/featured.json";
+
 const TOKEN_LIST_URL = 'https://raw.githubusercontent.com/balancer/tokenlists/main/generated/balancer.tokenlist.json';
 
+interface FeaturedPoolMetadata {
+    id: string;
+    imageUrl: string;
+    primary: boolean;
```
franzns, line 16: this is missing a chainId

mendesfabio:
hey I could add the filter by chains with support for param via headers. The query below seems to be returning what we needed - as specified in PR description: 

```gql
{
    "id": ID!,
    "primary": Boolean!,
    "imageUrl": String!,
    "pool": GqlPoolMinimal!
}
```

I didn't know exactly how to test the Sanity stuff and ensure the same query works there given `chains` param instead of `networkContext` - if you want to go ahead and fix that I appreciate.

image

modules/content/github-content.service.ts
```diff
@@ -158,8 +165,24 @@ export class GithubContentService implements ContentService {
         await prisma.prismaTokenType.createMany({ skipDuplicates: true, data: types });
     }
     async syncPoolContentData(): Promise<void> {}
-    async getFeaturedPoolGroups(): Promise<HomeScreenFeaturedPoolGroup[]> {
-        return [];
+    async getFeaturedPoolGroups(chains: Chain[]): Promise<HomeScreenFeaturedPoolGroup[]> {
+        const { data } = await axios.get<FeaturedPoolMetadata[]>(POOLS_METADATA_URL);
```
gmbronco, line 169: we have native fetch now, how about we move away from axios?

modules/pool/lib/pool-gql-loader.service.ts
```diff
@@ -154,8 +158,15 @@ export class PoolGqlLoaderService {
...
+    public async getFeaturedPoolGroups(chains: Chain[]): Promise<GqlPoolFeaturedPoolGroup[]> {
+        const featuredPoolGroups = [];
+        if (chains.some((chain) => BalancerChainIds.includes(chainToIdMap[chain]))) {
+            const githubContentService = new GithubContentService();
+            featuredPoolGroups.push(...(await githubContentService.getFeaturedPoolGroups(chains)));
+        } else if (chains.some((chain) => BeethovenChainIds.includes(chainToIdMap[chain]))) {
+            const sanityContentService = new SanityContentService('FANTOM');
+            featuredPoolGroups.push(...(await sanityContentService.getFeaturedPoolGroups(chains)));
+        }
         const poolIds = featuredPoolGroups
```
gmbronco, line 170: can we remove if / else and just query both? they will return empty array if there is no pools for a chain, right?

modules/content/github-content.service.ts
```diff
@@ -27,16 +36,14 @@ export class GithubContentService implements ContentService {
     async syncTokenContentData(): Promise<void> {
```
gmbronco, line 36: would it make sense to sync all the tokens at once without splitting per chain? i'm not sure if we aren't overcomplicating it here, since the whole list is fetched anyways.

gmbronco: bit lower there is upsert call, it can take chain as `token.chainId` instead of networkContext

modules/content/sanity-content.service.ts
```diff
@@ -77,12 +83,12 @@ export class SanityContentService implements ContentService {
 
             await prisma.prismaToken.upsert({
                 where: {
-                    address_chain: { address: tokenAddress, chain: networkContext.chain },
+                    address_chain: { address: tokenAddress, chain: this.chain },
```
gmbronco, line 86: isn't the chain returned in the response? is so we could use it here instead of setting it up externally

gmbronco (approved):

franzns (approved):

PR #518, opened 2023-11-07
Feature/multichain filter

franzns:
Adds chain and address filter to user and pool queries

Removes poolGetPoolFilters and poolGetUserSwapVolume query


modules/pool/pool.gql
```diff
@@ -1,5 +1,5 @@
 extend type Query {
-    poolGetPool(id: String!): GqlPoolBase!
+    poolGetPool(id: String!, chain: GqlChain): GqlPoolBase!
```
gmbronco, line 2: Do we need a chain filter when searching by ID? It's already unique, isn't it? same in poolGetSnapshots, poolJoinExit, userGetPoolJoinExits, userGetSwaps

franzns: at least it _should_ be unique, yes. No strong opinion, but looks like the safer way

PR #524, opened 2023-11-09
Onchain data fetching refactor

gmbronco:
* separating data fetching and parsing from database logic
* splitting into separate responsibilities (state, data, gyro)
* fixing protocol fee for pools in recovery - should be 0
* fixing BPT price rate


modules/pool/lib/pool-onchain-data.ts
```diff
@@ -0,0 +1,211 @@
...
+
+const defaultCalls = (
+    { id, address, type }: PoolInput,
+    vaultAddress: string,
+    multicaller: Multicaller3
+) => {
+    multicaller.call(`${id}.poolTokens`, vaultAddress, 'getPoolTokens', [id]);
+    multicaller.call(`${id}.totalSupply`, address, 'totalSupply');
+    multicaller.call(`${id}.virtualSupply`, address, 'getVirtualSupply');
+    multicaller.call(`${id}.actualSupply`, address, 'getActualSupply');
```
franzns: I would move these to the pool type specific queries, otherwise its an extra 2 queries per pool that will either fail or are not needed.

modules/pool/lib/pool-onchain-data.ts
```diff
@@ -0,0 +1,211 @@
...
+    multicaller.call(`${id}.swapFee`, address, getSwapFeeFn(type));
+    multicaller.call(`${id}.rate`, address, 'getRate');
+    multicaller.call(`${id}.protocolYieldFeePercentageCache`, address, 'getProtocolFeePercentageCache', [2]);
```
franzns, line 102: Same here, but I dont see a way to have this without relying on the network context :)

modules/pool/lib/pool-onchain-data.ts
```diff
@@ -0,0 +1,211 @@
...
+
+const stableCalls = (
+    { id, address, tokens }: PoolInput,
+    multicaller: Multicaller3
+) => {
+    multicaller.call(`${id}.amp`, address, 'getAmplificationParameter');
+
+    tokens.forEach(({ address: tokenAddress }, i) => {
+        multicaller.call(`${id}.tokenRate[${i}]`, address, 'getTokenRate', [tokenAddress]);
+        multicaller.call(`${id}.metaPriceRateCache[${i}]`, address, 'getPriceRateCache', [tokenAddress]);
```
franzns, line 131: same as above

modules/pool/lib/pool-onchain-data.ts
```diff
@@ -0,0 +1,211 @@
...
+
+const parse = (result: OnchainData, decimalsLookup: { [address: string]: number }) => ({
+    amp: result.amp ? formatFixed(result.amp[0], String(result.amp[2]).length - 1) : undefined,
+    swapFee: formatEther(result.swapFee ?? '0'),
+    totalShares: formatEther(result.actualSupply || result.virtualSupply || result.totalSupply || '0'),
+    weights: result.weights?.map(formatEther),
+    targets: result.targets?.map(String),
+    poolTokens: result.poolTokens ? {
+        tokens: result.poolTokens[0].map((t) => t.toLocaleLowerCase()),
+        balances: result.poolTokens[1].map((w, i) => formatUnits(w, decimalsLookup[result.poolTokens[0][i].toLocaleLowerCase()])),
```
franzns: can you change `toLocaleLowerCase()` to `toLowerCase()`? Just avoid any side effects

modules/pool/lib/pool-onchain-data.ts
```diff
@@ -0,0 +1,211 @@
...
+};
+
+const parse = (result: OnchainData, decimalsLookup: { [address: string]: number }) => ({
+    amp: result.amp ? formatFixed(result.amp[0], String(result.amp[2]).length - 1) : undefined,
+    swapFee: formatEther(result.swapFee ?? '0'),
+    totalShares: formatEther(result.actualSupply || result.virtualSupply || result.totalSupply || '0'),
+    weights: result.weights?.map(formatEther),
+    targets: result.targets?.map(String),
+    poolTokens: result.poolTokens ? {
+        tokens: result.poolTokens[0].map((t) => t.toLocaleLowerCase()),
```
franzns: I'm a fan of writing complete words for variables, we don't need to save any bytes :)

modules/pool/lib/pool-onchain-gyro-fee.ts
```diff
@@ -0,0 +1,86 @@
...
+    const multicaller = new Multicaller3(abi, batchSize);
+
+    const feeKey = formatBytes32String('PROTOCOL_SWAP_FEE_PERC');
+
+    const eclpKey = keccak256(
+        ['bytes'],
+        [
+            defaultAbiCoder.encode(
+                ['bytes32', 'bytes32'],
+                [feeKey, formatBytes32String('ECLP')]
```
franzns: Did you double check this key? any chance we can make this more general for other gyro pool types too?

modules/pool/lib/pool-on-chain-data.service.ts
```diff
@@ -90,37 +41,12 @@ export class PoolOnChainDataService {
...
-
-                if (multicallResult[pool.id] && multicallResult[pool.id].inRecoveryMode) {
-                    isInRecoveryMode = multicallResult[pool.id].inRecoveryMode;
-                }
-                if (multicallResult[pool.id] && multicallResult[pool.id].pausedState) {
-                    isPaused = multicallResult[pool.id].pausedState.paused;
-                }
                 operations.push(
                     prisma.prismaPoolDynamicData.update({
                         where: { id_chain: { id: pool.id, chain: networkContext.chain } },
```
franzns: any chance we could also get rid of the network context while we're refactoring that?

gmbronco: good point, moving it up to the pool service

modules/pool/lib/pool-onchain-data.ts
gmbronco: sure!

modules/pool/lib/pool-onchain-data.ts
gmbronco: good catch

modules/pool/lib/pool-onchain-data.ts
gmbronco: that's the part where we need to know the composable stable pool type, which depends on adding logic based on the factory address. i understand we don't need overfetching, but we fetch so much already that the cost is marginal. i tested both versions (this vs previous) and there is no perfomance impact.

modules/pool/lib/pool-onchain-gyro-fee.ts
gmbronco: confirmed as ECLP. yes, adding other types as well!

PR #530, opened 2023-11-15
V3 canary

franzns:
Changes:
- Chain filter for all pool queries
- Chain filter for all user queries
- Add balancerSubgraphService explicitly to all network configs to avoid network context dependency in subgraph service
 - introduction of services into network config
- Generalize APR group naming
- Removal of unused getUserSwapVolume query
- vebalVotingGauge service will run even if a voting gauge throws an error but still log error and fail the cron check. this makes sure that a wrongly configured voting gauge does not stop the complete sync.
- `poolGetPools` will also return blacklisted pools by default now.


modules/pool/pool.gql
```diff
@@ -1,12 +1,13 @@
...
+    poolGetPool(id: String!, chain: GqlChain): GqlPoolBase!
     poolGetPools(
         first: Int
         skip: Int
         orderBy: GqlPoolOrderBy
         orderDirection: GqlPoolOrderDirection
         where: GqlPoolFilter
         textSearch: String
+        userAddress: String
     ): [GqlPoolMinimal!]!
```
gmbronco: seems like it's a search param, but there is already `where`. is it ok to move it there?

modules/pool/pool.gql
```diff
@@ -15,16 +16,15 @@ extend type Query {
         orderDirection: GqlPoolOrderDirection
         where: GqlPoolFilter
         textSearch: String
+        userAddress: String
     ): Int!
```
gmbronco: same as above

modules/pool/pool.gql
franzns: fair

modules/pool/pool.gql
franzns: same as above

gmbronco:
there is a missed chain filter in: user-snapshot.service.ts#350 - https://github.com/beethovenxfi/beethovenx-backend/pull/530/files#diff-8125d1292097bfe85b74b26610b4e1b7c4714e7dfd25be08cb1f508c3e32253aR350

should be:
```
                where: {
                    id_chain: { id: poolId, chain },
                },
```

franzns:
> there is a missed chain filter in: user-snapshot.service.ts#350 - https://github.com/beethovenxfi/beethovenx-backend/pull/530/files#diff-8125d1292097bfe85b74b26610b4e1b7c4714e7dfd25be08cb1f508c3e32253aR350
> 
> should be:
> 
> ```
>                 where: {
>                     id_chain: { id: poolId, chain },
>                 },
> ```

there was even more in there, thanks!

PR #537, opened 2023-11-21
Feature/composable type

### LINKED ISSUE #529: add composableStablePool type
currently we only have phantomstablepool type which also includes composable stables. We should introduce a specific composableStablePool type which would help us getting rid of factory tracking.

This will have implication on the beets frontend, as it needs to know pool types for joins/exits

franzns:
addresses #529 

This will add composableStable type and also change phantom stable type to composableStable v0

need confirmation from @groninge01 before we can merge


modules/pool/lib/pool-on-chain-data.service.ts
```diff
@@ -95,12 +94,7 @@ export class PoolOnChainDataService {
         });
 
         const gyroPools = filteredPools.filter((pool) => pool.type.includes('GYRO'));
-        const poolsWithComposableStableType = filteredPools.map((pool) => ({
-            ...pool,
-            type: (isComposableStablePool(pool) ? 'COMPOSABLE_STABLE' : pool.type) as
-                | PrismaPoolType
-                | 'COMPOSABLE_STABLE',
-        }));
+        const poolsWithComposableStableType = filteredPools.filter((pool) => isComposableStablePool(pool));
```
gmbronco: the intention here was to assign COMPOSABLE_STABLE as a new type to know which contract calls to use to fetch data. eg: totalSupply. with the COMPOSABLE_STABLE in the DB we can just pass all the pools as they are.

```suggestion
```

gmbronco: i cannot suggest outside of edited lines. in [101](https://github.com/beethovenxfi/beethovenx-backend/pull/537/files#diff-a6ffa7676e335f304a2cad512cd8381718381d4bab532f16b221f6eb27ec8dceR101) line `poolsWithComposableStableType` needs to be `filteredPools`

gmbronco: nvm, just updated

gmbronco (approved):

modules/pool/lib/pool-on-chain-data.service.ts
franzns: 👍

PR #538, opened 2023-11-21
only have one set of resolvers and schema, keep folders for dev

### LINKED ISSUE #523: remove protocol specific resolvers
should remove the protocol specific resolvers and have the same schema for both beets and balancer

franzns:
addresses #523


gmbronco (approved):
lgtm

gmbronco:
btw. would it make sense to merge loadRestRoutesBeethoven + loadRestRoutesBalancer as well?

franzns:
> btw. would it make sense to merge loadRestRoutesBeethoven + loadRestRoutesBalancer as well?

I thought about it too, but it's not as important. I wanted to merge the schema because you *could* run into issues where a certain schema is deployed but not available. Can't happen with the restRoutes as they dont use codegen

That being said, might nevertheless make sense. I'm going to add it.

PR #549, opened 2023-12-01
unify poolType enum

### LINKED ISSUE #545: unify gqlPoolTypes
use enum GqlPoolType in all models:
- minimalPool
- GqlPoolTypeFilter
- all GqlPoolBase implementation

franzns:
addresses #545 

waiting for confirmation from @groninge01 before merging

also @agualis as it changes the type also in the vebal votinglist query


franzns:
maybe also affects b-sdk @brunoguerios ?

brunoguerios:
> maybe also affects b-sdk @brunoguerios ?

This doesn't affect `poolGetPool` query, right? It's the only one we're using so far.

PR #558, opened 2023-12-06
Refactor/get logs

gmbronco:
* ~~Adding Viem~~
* Isolating networkContext from UserSyncWalletBalanceService to see if that helps in tracing (👀 async context)
* Fetching events with array of addresses - so we can do one call instead of # of pool calls


gmbronco:
removed Viem from the PR

PR #588, opened 2023-12-14
faster logs fetching for staking

gmbronco:


modules/user/lib/user-sync-gauge-balance.service.ts
```diff
@@ -112,32 +108,63 @@ export class UserSyncGaugeBalanceService implements UserStakedBalanceService {
...
-                    address: gaugeAddress,
-                }),
-            );
-        }
-        const allResponses = await Promise.all(logPromises);
+        // Split the range into smaller chunks to avoid RPC limits, setting up to 50 times max block range
+        const maxBlockRange = networkContext.data.rpcMaxBlockRange;
+        const toBlock = Math.min(startBlock + 50 * maxBlockRange, latestBlock);
+        const range = toBlock - startBlock;
+        console.log(`user-sync-staked-balances-${networkContext.chainId} block range from ${startBlock} to ${toBlock}`);
```
franzns: Looks good, but would also isolate the networkContext as you did in the wallet sync service

gmbronco: done

PR #598, opened 2023-12-14
fix: missing Gyro type and data

gmbronco:
Some Gyro pools were added without a poolType and Gyro data. This PR assigns a proper type and upserts missing data on a pool syncing action.

### DIFF
--- a/modules/pool/lib/pool-creator.service.ts
+++ b/modules/pool/lib/pool-creator.service.ts
@@ -30,43 +30,72 @@ export class PoolCreatorService {
         for (const subgraphPool of sortedSubgraphPools) {
             console.log(`Syncing pool ${counter} of ${sortedSubgraphPools.length}`);
             counter = counter + 1;
-            const existsInDb = !!existingPools.find((pool) => pool.id === subgraphPool.id);
+            const pool = existingPools.find((pool) => pool.id === subgraphPool.id);
+            const existsInDb = !!pool;
 
             if (!existsInDb) {
                 await this.createPoolRecord(subgraphPool, blockNumber);
 
                 poolIds.push(subgraphPool.id);
             } else if (subgraphPool.poolType?.includes('Gyro')) {
-                await prisma.prismaPool.update({
-                    data: {
-                        gyroData: {
-                            update: {
-                                id: subgraphPool.id,
-                                alpha: subgraphPool.alpha || '',
-                                beta: subgraphPool.beta || '',
-                                sqrtAlpha: subgraphPool.sqrtAlpha || '',
-                                sqrtBeta: subgraphPool.sqrtBeta || '',
-                                root3Alpha: subgraphPool.root3Alpha || '',
-                                c: subgraphPool.c || '',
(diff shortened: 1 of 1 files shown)


modules/pool/lib/pool-creator.service.ts
```diff
@@ -30,44 +30,73 @@ export class PoolCreatorService {
...
             counter = counter + 1;
-            const existsInDb = !!existingPools.find((pool) => pool.id === subgraphPool.id);
+            const pool = existingPools.find((pool) => pool.id === subgraphPool.id);
+            const existsInDb = !!pool;
 
             if (!existsInDb) {
                 await this.createPoolRecord(subgraphPool, blockNumber);
 
                 poolIds.push(subgraphPool.id);
             } else if (subgraphPool.poolType?.includes('Gyro')) {
```
franzns, line 40: Instead of doing this, we should either handle that type update in the separate `reloadPoolTypes` mutation OR we change this `syncAllPoolsFromSubgraph` function that it would do all the updating (also versions, nested tokens, token indexes, etc)

franzns: The intention of this method is to only add new pools from subgraph that didnt make it into the DB via the `syncNewPoolsFromSubgraph` mutation. 
Maybe we should add an `syncAllPoolData` mutation that does all the re-syncing of versions, types, etc.

gmbronco: syncAllPoolData, or syncPoolData seems like a good idea, as it can consolidate all the subgraph-related tasks into one. It's fast anyways and it can simplify jobs a bit.

franzns: Some are not subgraph-related but I'd still add them to that new mutation

gmbronco: Initially, Gyro pools got stored before our DB schema was fully setup. We could sync pools only after they're supported. Imagine using something like `subgraphMappers` for different pool types. If a pool type doesn't have a mapper yet, we simply hold off on storing it until one is added. This would allow us to keep using syncAllPoolsFromSubgraph as usual, without the need to sync backwards.

However, if we're leaning towards backward syncing, we could consider combining these functions – syncAllPoolsFromSubgraph, poolSyncNewPoolsFromSubgraph, and syncPoolVersionForAllPools – into one, instead of adding another one. This could streamline things a bit.

franzns: I like the idea of the subgraphMappers though we wont be able to get rid of backward syncing. E.g. in the past we didn't have a pool version stored and later decided to store it so we had to add the backward sync for it. That would still be the case for if we had subgraphMappers.

I'd lean towards have a 'combined' function that does all the backwards syncing of types, versions, specific type fields, etc. into 'syncAllPoolData' or something. Maybe even bake the 'sync nested pools' and 'sync token indexes' into it.

gmbronco:
I opened a new PR #616 to work on subgraph mapping extraction.  I'm going to close this one, since all the pools are already synced properly.

PR #599, opened 2023-12-15
Refactor/sor service

franzns:
moving, renaming and removing unused stuff.
Also adding on chain query for v2 swaps


modules/sor/sor.gql
```diff
@@ -7,13 +7,41 @@ extend type Query {
...
+    swaps: [GqlSwap!]!
+    swapAmount: String!
+    swapAmountForSwaps: String!
+    returnAmount: String!
+    returnAmountFromSwaps: String!
+    returnAmountConsideringFees: String!
+    tokenIn: String!
+    tokenOut: String!
+    marketSp: String!
+}
```
gmbronco: perhaps we could already remove fields they don't use:

marketSp
returnAmountConsideringFees
swapAmountForSwaps

they didn't confirm which fields they use, didn't they?

modules/sor/sor.service.ts
```diff
@@ -1,54 +1,81 @@
...
-        const swap = await this.getSwap({ ...input, swapOptions: {} });
-        const emptyResponse = EMPTY_COWSWAP_RESPONSE(input.tokenIn, input.tokenOut, input.swapAmount);
+    async getCowSwaps(args: QuerySorGetCowSwapsArgs): Promise<GqlCowSwapApiResponse> {
+        console.log('getCowSwaps args', JSON.stringify(args));
+        const amountToken = args.swapType === 'EXACT_IN' ? args.tokenIn : args.tokenOut;
+        // Use TokenAmount to help follow scaling requirements in later logic
+        // args.swapAmount is RawScale, e.g. 1USDC should be passed as 1000000
+        const amount = await getTokenAmountRaw(amountToken, args.swapAmount, args.chain!);
+
+        const swap = await sorV2Service.getSwapResult({
```
gmbronco, line 25: V2 only?

modules/sor/sor.service.ts
```diff
@@ -1,54 +1,81 @@
...
+    async getSorSwaps(args: QuerySorGetSwapsArgs): Promise<GqlSorGetSwapsResponse> {
+        console.log('getSorSwaps args', JSON.stringify(args));
+        const tokenIn = args.tokenIn.toLowerCase();
+        const tokenOut = args.tokenOut.toLowerCase();
+        const amountToken = args.swapType === 'EXACT_IN' ? tokenIn : tokenOut;
+        // Use TokenAmount to help follow scaling requirements in later logic
+        // args.swapAmount is HumanScale
+        const amount = await getTokenAmountHuman(amountToken, args.swapAmount, args.chain!);
+
+        const swap = await this.getComparingSwap({
```
gmbronco, line 55: i like the name vs getSwap

gmbronco (approved):

modules/sor/sor.gql
franzns: agree, will remove once I know what they use

modules/sor/sor.service.ts
franzns: yes, balancer chains dont support a V1

PR #600, opened 2023-12-15
V3 canary

franzns:


modules/beethoven/balancer-sor.service.ts
```diff
@@ -331,7 +332,7 @@ export class BalancerSorService {
     }
 
     private getTokenDecimals(tokenAddress: string, tokens: PrismaToken[]): number {
-        if (tokenAddress === '0x0000000000000000000000000000000000000000') {
+        if (tokenAddress === ZERO_ADDRESS || tokenAddress === NATIVE_ADDRESS) {
```
gmbronco, line 335: @franzns – would we need to add 0x00...1010 here as well?

PR #602, opened 2023-12-15
sync-latest-fx-prices

### LINKED ISSUE #565: SOR: Handle failing FX pools by adding latestFXPrice to it's poolTokens.
- [x] Find out if we need FX pools. - we do!
- [x] FX pool tokens are missing latestFXPrice - subgraph has it and it needs to be added to the DB as well.
- [x] Once done uncomment FX pool from the query filter: https://github.com/beethovenxfi/beethovenx-backend/blob/v3-canary/modules/sor/sorV2/sorV2.service.ts/#L361

gmbronco:
Adding a job for syncing latest FX price from subgraph. The task itself is fast, so I'm scheduling it as one job for all the chains.

closes #565

### DIFF
--- a/modules/network/optimism.ts
+++ b/modules/network/optimism.ts
@@ -298,7 +298,10 @@ export const optimismNetworkConfig: NetworkConfig = {
     ],
     userStakedBalanceServices: [new UserSyncGaugeBalanceService()],
     services: {
-        balancerSubgraphService: new BalancerSubgraphService(optimismNetworkData.subgraphs.balancer, optimismNetworkData.chain.id),
+        balancerSubgraphService: new BalancerSubgraphService(
+            optimismNetworkData.subgraphs.balancer,
+            optimismNetworkData.chain.id,
+        ),
     },
     /*
     For sub-minute jobs we set the alarmEvaluationPeriod and alarmDatapointsToAlarm to 1 instead of the default 3. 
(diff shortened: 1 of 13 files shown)


franzns:
As discussed, I would refrain from introducing the "multichain-schedule" for this. Would like a more generalized approach.

gmbronco:
Removed multichain job in favour of chain specific ones.

modules/network/optimism.ts
```diff
@@ -406,5 +409,9 @@ export const optimismNetworkConfig: NetworkConfig = {
             name: 'feed-data-to-datastudio',
             interval: (env.DEPLOYMENT_ENV as DeploymentEnv) === 'canary' ? every(5, 'minutes') : every(1, 'minutes'),
         },
+        {
+            name: 'sync-latest-fx-prices',
+            interval: every(10, 'minutes'),
+        },
```
franzns: we dont have fx pools on OP. not sure about the other chains

modules/network/zkevm.ts
```diff
@@ -292,5 +292,9 @@ export const zkevmNetworkConfig: NetworkConfig = {
             name: 'feed-data-to-datastudio',
             interval: (env.DEPLOYMENT_ENV as DeploymentEnv) === 'canary' ? every(5, 'minutes') : every(1, 'minutes'),
         },
+        {
+            name: 'sync-latest-fx-prices',
+            interval: every(10, 'minutes'),
+        },
```
franzns: same, think they dont have fx pools

gmbronco: true, FX is on: Mainnet, Polygon and Avalanche so far.

PR #606, opened 2023-12-18
Handle RPC range errors in getLogs

gmbronco:
Extracted getLogs into separate function where we can handle RPC errors related to maximum block range thrown differently by different providers. Added an option to decode event data when ABI is passed.

### DIFF
--- a/modules/user/lib/user-sync-gauge-balance.service.ts
+++ b/modules/user/lib/user-sync-gauge-balance.service.ts
@@ -139,8 +139,8 @@ export class UserSyncGaugeBalanceService implements UserStakedBalanceService {
             therefore we check all transfer events since the last synced block
          */
 
-        // Split the range into smaller chunks to avoid RPC limits, setting up to 50 times max block range
-        const toBlock = Math.min(startBlock + 50 * this.rpcMaxBlockRange, latestBlock);
+        // Split the range into smaller chunks to avoid RPC limits, setting up to 5 times max block range
+        const toBlock = Math.min(startBlock + 5 * this.rpcMaxBlockRange, latestBlock);
         console.log(`user-sync-staked-balances-${this.chainId} block range from ${startBlock} to ${toBlock}`);
         console.log(`user-sync-staked-balances-${this.chainId} getLogs for ${gaugeAddresses.length} gauges.`);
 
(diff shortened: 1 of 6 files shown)


modules/user/lib/user-sync-gauge-balance.service.ts
```diff
@@ -123,7 +123,9 @@ export class UserSyncGaugeBalanceService implements UserStakedBalanceService {
                 select: { gaugeAddress: true },
                 where: { chain: this.chain },
             })
-        ).map((gauge) => gauge.gaugeAddress);
+        )
+            .map((gauge) => gauge.gaugeAddress)
+            .filter((address, i, arr) => arr.indexOf(address) === i); // Make unique
```
franzns: they should already be unique?

gmbronco: true, good catch, removing.

modules/web3/events.ts
```diff
@@ -0,0 +1,120 @@
...
+                        // Infura returns 'more than 10000 results' error if block range is too wide
+                        if (e.includes('query returned more than 10000 results')) {
+                            const range = e
+                                .match(/\[([0-9a-fA-F, x]+)\]/)
+                                .pop()
+                                .split(', ')
+                                .map((hex: string) => parseInt(hex, 16));
+
+                            return getEvents(from, to, addresses, topics, rpcUrl, range[1] - range[0]);
+                        }
```
franzns: These retries will mess with the "latestSyncBlock" logic. If it retries, it will still persist the toBlock as the latest think block and not the one it actually synced to. You either need to make sure all blocks up until toBlock are queried or get rid of the retries and set reasonable block limits.

I guess we anyway need to make sure that it fails if one of the promises fails, or not do promise.all but do they sequentially and return the block number until which we synced

gmbronco: thanks, I double checked and unless I'm missing something obvious, catch is refetching only the chunk that failed, not the whole range, so it's just filling the gaps. there is promise.all instead of allSettled to make sure it gets rejected as a whole, not just a part of it.

calling getEvents should return the whole range or nothing.

modules/web3/events.ts
```diff
@@ -0,0 +1,131 @@
...
+
+                // Infura returns 'more than 10000 results' error if block range is too wide
+                if (e.includes('query returned more than 10000 results')) {
+                    const range = e
+                        .match(/\[([0-9a-fA-F, x]+)\]/)
+                        .pop()
+                        .split(', ')
+                        .map((hex: string) => parseInt(hex, 16));
+
+                    return getEvents(from, to, addressChunk, topics, rpcUrl, range[1] - range[0]);
```
franzns: Could you describe the actual string you are parsing here in the comments? Hard to say whether the string manipulation is correct

franzns (approved):

PR #631, opened 2024-01-09
V3 canary

franzns:
- add sentry integrations
- optimize lifetime value update
- set cron intervals to standard on three balancer chains (mainnet, polygon, arb)
- run purge token prices only once for all chains

### DIFF
--- a/app.ts
+++ b/app.ts
@@ -27,7 +27,6 @@ async function startServer() {
 
     Sentry.init({
         dsn: env.SENTRY_DSN,
-        // tracesSampleRate: 0.005,
         environment: `multichain-${env.DEPLOYMENT_ENV}`,
         enabled: env.NODE_ENV === 'production',
         ignoreErrors: [/.*error: Provide.*chain.*param/],
@@ -39,8 +38,8 @@ async function startServer() {
             new Sentry.Integrations.Http({ tracing: true }),
             new ProfilingIntegration(),
         ],
-        tracesSampleRate: 0.2,
-        profilesSampleRate: 0.1,
+        tracesSampleRate: 0.005,
+        profilesSampleRate: 0.005,
         beforeSend(event, hint) {
             const error = hint.originalException as string;
             if (error?.toString().includes('Unknown token:')) {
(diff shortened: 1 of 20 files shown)


app.ts
```diff
@@ -20,26 +20,26 @@ import { ProfilingIntegration } from '@sentry/profiling-node';
...
+            new Sentry.Integrations.GraphQL(),
+            new Sentry.Integrations.Prisma({ client: prisma }),
             new Sentry.Integrations.Express({ app }),
             new Sentry.Integrations.Http({ tracing: true }),
             new ProfilingIntegration(),
         ],
-        tracesSampleRate: 0.2,
-        profilesSampleRate: 0.1,
+        tracesSampleRate: 0.005,
+        profilesSampleRate: 0.005,
```
gmbronco: this one depends on tracesSampleRate, so now it's 0.005 of 0.005 total traces. we can leave it as 0.1 - 10% of traces.

modules/token/token.service.ts
```diff
@@ -216,8 +216,8 @@ export class TokenService {
         return this.tokenPriceService.getHistoricalTokenPrices();
     }
 
-    public async purgeOldTokenPrices(): Promise<number> {
-        return this.tokenPriceService.purgeOldTokenPrices();
+    public async purgeOldTokenPricesForAllChains() {
+        await this.tokenPriceService.purgeOldTokenPricesForAllChains();
```
gmbronco: async / await can be skipped here, because we don't use results within the function and  it returns a promise anyways

worker/worker.ts
```diff
@@ -12,12 +13,15 @@ export async function startWorker() {
...
+            new Sentry.Integrations.GraphQL(),
+            new Sentry.Integrations.Prisma({ client: prisma }),
+            new Sentry.Integrations.Express({ app }),
             new Sentry.Integrations.Http({ tracing: true }),
             new ProfilingIntegration(),
         ],
-        tracesSampleRate: 0.2,
-        profilesSampleRate: 0.1,
+        tracesSampleRate: 0.005,
+        profilesSampleRate: 0.005,
```
gmbronco: same as in the app

gmbronco (approved):

